PR #8, opened 2018-11-29
Trying out ethers.js with React Native

marcelomorgado:


demo/App.js
```diff
@@ -1,12 +1,177 @@
...
 
 export default class App extends React.Component {
   state = {
     isLoadingComplete: false
   };
 
+
+	componentDidMount() {
+
+		let ropstenProvider = ethers.getDefaultProvider('ropsten');
```
pcowgill: @marcelomorgado I think this would be better as a `const` instead of a `let`, since it doesn't change.

demo/App.js
```diff
@@ -1,12 +1,177 @@
...
+		let fundedWallet = new ethers.Wallet("0x18bfdd05c1e63e4a27081352e2910c164a4d34588f8d7eecfdfcb654742934c2", ropstenProvider); //addr:0xa9adeF68CEF4EA3C9d0717126e14fDfebFDFB4d7
+
+		let wallet2mnemonic = "radar blur cabbage chef fix engine embark joy scheme fiction master release";
+		let wallet2 = ethers.Wallet.fromMnemonic(wallet2mnemonic);
+		let wallet3 = ethers.Wallet.createRandom();
+
+		console.assert(wallet.address === address, "wallet.address");
+		console.assert(wallet.privateKey === privateKey, "wallet.privateKey");
+		console.assert(wallet2.mnemonic === wallet2mnemonic, "wallet.privateKey");
+		console.assert(wallet3.address !== undefined, "wallet.address from randonly created");
```
pcowgill: @marcelomorgado Do you mean `rand only` or `randomly`?

demo/App.js
pcowgill:
```suggestion
		const ropstenProvider = ethers.getDefaultProvider('ropsten');
```

demo/App.js
```diff
@@ -1,12 +1,177 @@
...
   };
 
+
+	componentDidMount() {
+
+		let ropstenProvider = ethers.getDefaultProvider('ropsten');
+		let mainnetProvider = ethers.getDefaultProvider();
+		let utils = ethers.utils;
+
+		let privateKey = "0x3141592653589793238462643383279502884197169399375105820974944592";
```
pcowgill:
```suggestion
		const privateKey = "0x3141592653589793238462643383279502884197169399375105820974944592";
```

demo/App.js
```diff
@@ -1,12 +1,177 @@
...
+		let address = "0x7357589f8e367c2C31F51242fB77B350A11830F3";
```
pcowgill:
```suggestion
		const address = "0x7357589f8e367c2C31F51242fB77B350A11830F3";
```

demo/App.js
```diff
@@ -1,12 +1,177 @@
...
+		//             iv: "1dcdf13e49cea706994ed38804f6d171"
+		//          }
+		//     },
+		//     "version" : 3
+		// };
+		//
+		// let json = JSON.stringify(data);
+		// let password = "foo";
+		//
+		// ethers.Wallet.fromEncryptedWallet(json, password).then(function(wallet) {
```
pcowgill: @marcelomorgado Is this from the ethers.js docs? Either way, let's aim to stick with async/await syntax wherever possible. Thanks!

demo/App.js
```diff
@@ -1,12 +1,177 @@
...
+			nonce: 0,
+			gasLimit: 21000,
+			gasPrice: ethers.utils.bigNumberify("20000000000"),
+			to: "0x88a5C2d9919e46F883EB62F7b8Dd9d0CC45bc290",
+			value: ethers.utils.parseEther("1.0"),
+			data: "0x",
+			chainId: ethers.utils.getNetwork('homestead').chainId
+		}
+
+		fundedWallet.sign(transaction).then((signedTransaction) => {
```
pcowgill: @marcelomorgado Let's change this to async/await before merging, if that works for you

demo/App.js
```diff
@@ -1,12 +1,177 @@
 import React from "react";
 import { StyleSheet, Text, View } from "react-native";
 import { AppLoading, Asset, Font } from "expo";
+import "ethers/dist/shims.js";
+import { ethers } from "ethers";
```
pcowgill:
```suggestion
import { ethers } from "ethers";

// Note: In the future, we'll be importing the Tasit SDK
// which has ethers.js as a dependency rather than importing
// it directly
```

demo/App.js
marcelomorgado: Ok

demo/App.js
marcelomorgado: randomly. Sorry

demo/App.js
marcelomorgado: [Yes](https://docs.ethers.io/ethers.js/html/api-wallet.html#encrypted-json-wallets). Got it.

demo/App.js
marcelomorgado: Sure

PR #10, opened 2018-12-01
SDK dependency

pcowgill:


marcelomorgado:
Dependency works with local dir npm installation. That can be finished when tasit-sdk npm package been published.

demo/App.js
```diff
@@ -1,12 +1,19 @@
...
 import { AppLoading, Asset, Font } from "expo";
+//import { Account } from "tasit-sdk";
 
 export default class App extends React.Component {
   state = {
     isLoadingComplete: false
   };
 
+  componentDidMount() {
+    // Account.create().then(w => {
```
pcowgill: @marcelomorgado Even though this isn't being used since it's commented out and the dep isn't live on npm yet, I'd  still appreciate it if we could swap it to async/await style. Thanks!

pcowgill:
@marcelomorgado So you're thinking once the npm package has been published we'll add `tasit-sdk` as a dependency in package.json?

marcelomorgado:
Yes. With the latest change at [features/rn-compatibily](https://github.com/tasitlabs/TasitSDK/pull/28), `tasit-sdk` is importing ethers.js shims correctly.

pcowgill:
@marcelomorgado The `tasit-sdk` can be added as a dep now that it's published https://github.com/tasitlabs/TasitSDK/issues/30

marcelomorgado:
Demo app has now a sample of using TasitSDK.Account.

demo/App.js
marcelomorgado: OK

PR #13, opened 2018-12-03
Lint using prettier + eslint

marcelomorgado:
These rules were needed to avoid errors on current App.js file:
``
"no-console": 0, 
"react/prop-types": 0,
``
I'm not sure about the best way how `npm run lint` should be, I've saw some examples that style/code script were splitted but I thought that is a bit verbose. What do you think?
Another point is about how files shoud be under linting. For now it's applying under App.js only.

### DIFF
--- a/demo/.eslintrc.js
+++ b/demo/.eslintrc.js
@@ -9,6 +9,7 @@ module.exports = {
 	extends: [
 		"eslint:recommended",
 		"plugin:react/recommended",
+		"plugin:react-native/all",
 		"prettier",
 		"prettier/react",
 	],
--- a/demo/.eslintrc.js
+++ b/demo/.eslintrc.js
@@ -1,0 +1,24 @@
+module.exports = {
+	parser: "babel-eslint",
+	env: {
+		browser: true,
+		es6: true,
+	},
+	plugins: ["react", "prettier", "react-native"],
+	extends: [
+		"eslint:recommended",
+		"plugin:react/recommended",
+		"prettier",
+		"prettier/react",
+	],
+	rules: {
+		"no-console": 0,
+		"react/prop-types": 0,
+		"prettier/prettier": "error",
+	},
+	settings: {
+		react: {
+			version: "16.5",
+		},
+	},
+};
--- a/demo/.eslintrc.js
+++ b/demo/.eslintrc.js
@@ -3,6 +3,7 @@ module.exports = {
 	env: {
 		browser: true,
 		es6: true,
+		node: true,
 	},
 	plugins: ["react", "prettier", "react-native"],
 	extends: [
--- a/demo/package.json
+++ b/demo/package.json
@@ -10,7 +10,7 @@
     "ios": "npx expo start --ios",
     "eject": "npx expo eject",
     "test": "npx jest",
-		"lint": "npx prettier --write '{*.js,**/*.js,*.jsx,**/*.jsx}' && npx eslint '{*.js,**/*.js,*.jsx,**/*.jsx}'"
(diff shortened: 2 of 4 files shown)


demo/package.json
```diff
@@ -9,7 +9,8 @@
     "android": "npx expo start --android",
     "ios": "npx expo start --ios",
     "eject": "npx expo eject",
-    "test": "npx jest"
+    "test": "npx jest",
+		"lint": "npx prettier --write App.js && npx eslint App.js"
```
pcowgill: The indentation is weird compared with the rest of the file

pcowgill:
We'll have a large directory of React Native components that we'll want linted too. I think basically any file ending in .js or .jsx should be linted. Does that make sense to you @marcelomorgado ?

pcowgill:
If I remember correctly eslint now prefers words as opposed to numbers for their rule settings. We don't necessarily need to change the rules to match the current App.js - we could modify App.js as well. As long as the current settings still produce warnings though we can track that and work on it in a future PR.

pcowgill:
Once we have CI configured I think it makes sense to run linting and tests as separate scripts that can fail independently and linking out to the results from GitHub PRs. But until we have that set up I think having lint run automatically when you run test is good.

demo/package.json
```diff
@@ -20,7 +21,13 @@
     "tasit-sdk": "0.0.2"
   },
   "devDependencies": {
+    "babel-eslint": "^10.0.1",
     "babel-preset-expo": "^5.0.0",
+    "eslint": "^5.9.0",
+    "eslint-config-prettier": "^3.3.0",
+    "eslint-plugin-prettier": "^3.0.0",
```
pcowgill: @marcelomorgado Are both the `eslint-config-prettier` and `eslint-plugin-prettier` required? They may very well be - I forget what the distinction is.

demo/package.json
pcowgill:
```suggestion
    "lint": "npx prettier --write App.js && npx eslint App.js"
```

marcelomorgado:
> We'll have a large directory of React Native components that we'll want linted too. I think basically any file ending in .js or .jsx should be linted. Does that make sense to you @marcelomorgado ?

Yes, I'll do that.

demo/.eslintrc.js
```diff
@@ -0,0 +1,25 @@
...
+	parser: "babel-eslint",
+	env: {
+		browser: true,
+		es6: true,
+		node: true,
+	},
+	plugins: ["react", "prettier", "react-native"],
+	extends: [
+		"eslint:recommended",
+		"plugin:react/recommended",
```
pcowgill:
```suggestion
		"plugin:react/recommended",
		"plugin:react-native/all",
```

demo/package.json
```diff
@@ -9,7 +9,8 @@
     "android": "npx expo start --android",
     "ios": "npx expo start --ios",
     "eject": "npx expo eject",
-    "test": "npx jest"
+    "test": "npx jest",
+		"lint": "npx prettier --write '{*.js,**/*.js,*.jsx,**/*.jsx}' && npx eslint '{*.js,**/*.js,*.jsx,**/*.jsx}'"
```
pcowgill:
```suggestion
    "lint": "npx prettier --write '{*.js,**/*.js,*.jsx,**/*.jsx}' && npx eslint '{*.js,**/*.js,*.jsx,**/*.jsx}'"
```

demo/package.json
marcelomorgado: @pcowgill `eslint-plugin-prettier` is used to run `prettier` by `eslint` and uses its as rules.
`eslint-config-prettier` avoid conflicts between both disabling `eslint` formatting rules.
More info: https://github.com/prettier/eslint-plugin-prettier#recommended-configuration

Issue #19, opened 2018-12-07
Remaining small issues from PR 17


pcowgill:
I merged https://github.com/tasitlabs/tasit/pull/17, so I created this issue to track the remaining tiny conversation threads.
 
https://github.com/tasitlabs/tasit/pull/17#discussion_r239838634 (and https://github.com/tasitlabs/tasit/pull/17#discussion_r239904745)
https://github.com/tasitlabs/tasit/pull/17#discussion_r239904263
https://github.com/tasitlabs/tasit/pull/17#discussion_r239573789

PR #26, opened 2018-12-17
Remaining modifications from an earlier PR

marcelomorgado:
Refs #19

### DIFF
--- a/demo/.eslintrc.js
+++ b/demo/.eslintrc.js
@@ -14,8 +14,9 @@ module.exports = {
     "prettier/react",
   ],
   rules: {
-    "no-console": 0,
-    "react/prop-types": 0,
+    "no-console": "off",
+    "react-native/no-raw-text": "off",
+    "react/prop-types": "off",
     "prettier/prettier": "error",
   },
   settings: {
--- a/demo/components/presentational/LargeText.js
+++ b/demo/components/presentational/LargeText.js
@@ -1,0 +1,18 @@
+import React from "react";
+import { StyleSheet, Text } from "react-native";
+import { responsiveHeight } from "react-native-responsive-dimensions";
+import Colors from "../../constants/Colors";
+
+export default function LargeText(props) {
+  return <Text style={styles.text}>{props.children}</Text>;
+}
+
+const styles = StyleSheet.create({
+  text: {
+    padding: responsiveHeight(2),
+    fontSize: 30,
+    textAlign: "center",
+    fontWeight: "bold",
+    color: Colors.textColor,
+  },
+});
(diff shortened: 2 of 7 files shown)


demo/.eslintrc.js
```diff
@@ -14,8 +14,9 @@ module.exports = {
     "prettier/react",
   ],
   rules: {
-    "no-console": 0,
-    "react/prop-types": 0,
+    "no-console": "off",
+    "react-native/no-raw-text": "off",
+    "react/prop-types": "off",
```
pcowgill: I think we want `react/prop-types` enabled in the long run. But either way, this change to strings for the config rather than integers is a great change!

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md

demo/.eslintrc.js
```diff
@@ -14,8 +14,9 @@ module.exports = {
     "prettier/react",
   ],
   rules: {
-    "no-console": 0,
-    "react/prop-types": 0,
+    "no-console": "off",
+    "react-native/no-raw-text": "off",
```
pcowgill: I think we want `react-native/no-raw-text` enabled in the long run. But either way, this change to strings for the config rather than integers is a great change!

https://github.com/Intellicode/eslint-plugin-react-native/blob/master/docs/rules/no-raw-text.md

demo/.eslintrc.js
```diff
@@ -14,8 +14,9 @@ module.exports = {
     "prettier/react",
   ],
   rules: {
-    "no-console": 0,
-    "react/prop-types": 0,
+    "no-console": "off",
```
pcowgill: I think in the long run we want `no-console` on or at least allowing `console.warn` and `console.error`. If we have it on, we can make a function that wraps the console.log functions and only uses them when a dev env var is set to true or something like that. But either way, this change to strings for the config rather than integers is a great change!


https://eslint.org/docs/rules/no-console

demo/components/presentational/LargeText.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import { StyleSheet, Text } from "react-native";
+import { responsiveHeight } from "react-native-responsive-dimensions";
+import Colors from "../../constants/Colors";
```
pcowgill: I created an issue for myself to add a name in package.json for these https://github.com/tasitlabs/tasit/issues/27

demo/.eslintrc.js
pcowgill: This can wait until a future issue, though https://github.com/tasitlabs/tasit/issues/28

demo/.eslintrc.js
pcowgill: This can wait until a future issue, though https://github.com/tasitlabs/tasit/issues/28

Issue #12, opened 2018-12-03
Add a test to the test suite confirming that Account.create() works in React Native env


pcowgill:

pcowgill:
This makes sure that the shims we used with `ethers.js` over in the Tasit SDK made it through to here properly and are working in React Native now that there's this indirection in where they come from.

pcowgill:
@marcelomorgado Thanks for this detailed info! That test case is what I was picturing.

I think in general having as many shallow tests as possible is considered good practice.

For most smaller components that change based on variable data, we can test those just by faking that data as input to a pure component. So we won't have too many tests where we need to manually call componentDidMount to trigger data generation.

Enzyme is what I've used in the past when I do need to manually interact with the virtual DOM.

So let's mostly use the non-Enzyme approach, but use Enzyme for this specific case you brought up. (So that would mean using a little of both open WIP branches).

In the Enzyme README they say Enzyme v3 is supposed to support React 16.

I followed through that path of comments you linked to. Let's create an issue to remember to remove the modifiied preprocessor whenever they get this working.

I haven't encountered this particular issue with react-native before, mostly because I've mostly worked with older versions.

Here are some slides I like on testing React (and React Native) applications https://docs.google.com/presentation/d/1VRyWAZ0qSmJgcCEpl7XX68eoVf7890H1nkPWM0jkToY/edit#slide=id.p5

pcowgill:
@marcelomorgado Given that info, which WIP will you plan on working on?

marcelomorgado:
Snapshots

I’ve found three ways to do that:
1. Using react-test-renderer create
2. Using react-test-renderer/shallow
3. Using enzyme shallow

I’ve used Jest+Enzyme to generate snapshots. **Motivations:**
- The attached presentation uses it;
- react-test-renderer (renderer.create) is rendering component entirely creating a large snapshot (there is a lint rule to prevent large snapshots) and testing child components;
- The Enzyme has a better shallow API.

It isn’t working perfectly yet. There is a caveat:

Snapshots don’t recognize some react-native components (i.e., View, Image are saved as Component)?
**Issue**: https://github.com/facebook/react-native/pull/22417
**Status**: Solved on react-native v0.57.8 (2018/12/13) [release note](https://github.com/facebook/react-native/releases/tag/v0.57.8)
**Solution**: Wait for new Expo release with RN v0.57.8. Current `expo` package is 31.0.6 (2018/11/30)

**Refs**
[...]

Issue #14, opened 2018-12-03
Update the linting rules and update the components to adhere to the new rules


marcelomorgado:
https://github.com/tasitlabs/tasit/pull/13#issuecomment-443866741

How can we do about current [console.log](https://github.com/tasitlabs/tasit/blob/develop/demo/App.js#L13). Could we display wallet address on screen? Any better idea?

pcowgill:
Yeah, displaying it on the screen seems like a good plan!

marcelomorgado:
Also upgrade eslint rules config to use words instead numbers.

pcowgill:
@marcelomorgado I think this could be a good quick one to knock out soon

pcowgill:
A few other thoughts on the linting settings for this issue:
https://github.com/tasitlabs/tasit/pull/26#discussion_r242397281
https://github.com/tasitlabs/tasit/pull/26#discussion_r242397420
https://github.com/tasitlabs/tasit/pull/26#discussion_r242397756

marcelomorgado:
**- no-console**
There is one use of it now: https://github.com/tasitlabs/tasit/blob/aff489b2fb4d41cf15cd694840960bb49243708b/demo/App.js#L31
Can I comment that?

**- react-native/no-raw-text**
I've turned off this rule because I wanted that `<LargeText>` component behavior was as `<Text>`. Is be better with this will be turned on? If is the case, the large text component will be used that way:
```
<LargeText><Text>String</Text></LargeText>
```

**- jest/no-large-snapshots**
I've activated that rule because I've thought it interesting to help us to keep the components small. But seems that 20 lines could be too small. 6/7 snapshots are grater than 20 lines.

pcowgill:
**no-console**
I'd slightly prefer to change the config for the no-console rule to allow warn and error for now. https://eslint.org/docs/rules/no-console

In the long run, we can make a logging helper that wraps the console.log / console.warn / etc. functions and only uses them when a dev env var is set to true.

pcowgill:
**jest/no-large-snapshots**

Yes, let's change it to 40 or 50. How does that sound? Would the snapshots pass with that new rule?

pcowgill:
**react-native/no-raw-text**
```
This rule has an object option:

    "skip" – allow to skip checking for the array of custom components
```

I think that should solve the problem!

pcowgill:
[...]

PR #29, opened 2018-12-18
First round of tests for the demo mobile app

marcelomorgado:
Refs #12 

Snapshots: https://github.com/tasitlabs/tasit/issues/12#issuecomment-448298933
Enzyme: https://github.com/tasitlabs/tasit/issues/12#issuecomment-448300733


demo/components/presentational/EthereumSignUpForm.js
```diff
@@ -1,9 +1,17 @@
 import React from "react";
 import { Button, StyleSheet, View, Text, TextInput } from "react-native";
+import { Account } from "tasit-sdk";
 
 export default class EthereumSignUpForm extends React.Component {
   state = {
     text: "",
+    address: "",
```
pcowgill: This is a good candidate for moving to redux in the future

demo/components/presentational/EthereumSignUpForm.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+  it("renders the component", async () => {
+    expect(shallow(<EthereumSignUpForm />)).toMatchSnapshot();
+  });
+
+  it("creates a wallet - calling function", async () => {
+    const wrapper = shallow(<EthereumSignUpForm />);
+    expect(wrapper.state("address")).toEqual("");
+    await wrapper.instance().createAccount();
+    expect(wrapper.state("address")).not.toEqual("");
```
pcowgill: We could assert that the address is the right length here too. But maybe the Tasit SDK tests will already cover that sufficiently.

demo/jest.preprocessor.js
```diff
@@ -0,0 +1,124 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow
+ */
+
+/* eslint-env node */
+
+"use strict";
+
```
pcowgill: Should we create an issue and/or add a TODO in the code to remind us of the conditions under which we can remove this file?

demo/package.json
```diff
@@ -10,7 +10,7 @@
     "ios": "npx expo start --ios",
     "eject": "npx expo eject",
     "test": "npx jest",
-    "lint": "npx prettier --write '{*.js,**/*.js,*.jsx,**/*.jsx}' && npx eslint '{*.js,**/*.js,*.jsx,**/*.jsx}'"
+    "lint": "npx prettier --write '{*.js,**/*.js,*.jsx,**/*.jsx}' && npx eslint '{*.js,**/*.js,*.jsx,**/*.jsx,**/*.snap}'"
```
pcowgill: Can you expand on the choice of linting the snapshots? I think it seems reasonable, but I haven't given this much thought before.

demo/package.json
```diff
@@ -23,22 +23,39 @@
...
-    "preset": "jest-expo"
+    "preset": "jest-expo",
+    "transform": {
+      "\\.js$": "<rootDir>/jest.preprocessor.js"
+    },
+    "setupFiles": [
+      "<rootDir>/jest.setup.js"
+    ],
+    "transformIgnorePatterns": [
+      "node_modules/(?!(jest-)?react-native|.*react-navigation|expo-react-native-adapter)"
```
pcowgill: What's the goal of this regex?

demo/screens/Home.js
```diff
@@ -1,16 +1,9 @@
 import React from "react";
 import { Image, Button, StyleSheet, View } from "react-native";
-import { Account } from "tasit-sdk";
 import LargeText from "../components/presentational/LargeText";
 import Colors from "../constants/Colors";
 
 export default class Home extends React.Component {
-  componentDidMount = async () => {
```
pcowgill: I like that you moved this to the signup component

demo/jest.preprocessor.js
marcelomorgado: Yes, sounds good.

demo/package.json
marcelomorgado: I'm not sure if it is a good idea but I saw an eslint rule to keep snapshots small. https://github.com/tasitlabs/tasit/issues/14#issuecomment-448313550
What do you think?

demo/jest.preprocessor.js
pcowgill: Done https://github.com/tasitlabs/tasit/issues/31

demo/package.json
pcowgill: Ahhhh got it. I definitely wanted to apply the rule to keep snapshots small - it just didn't cross my mind when reading this line that this change is what would make that possible. Sounds good!

demo/package.json
pcowgill: That's a great explanation. Thanks!

Issue #43, opened 2018-12-19
Use enzyme button click for the the account creation test after updating the SDK


pcowgill:

marcelomorgado:
Actually, that isn't related to Expo SDK version.

**Problem:**
Enzyme.simulate('click') doesn't support `async handlers` (and doesn't will).
Refs: https://github.com/airbnb/enzyme/pull/867#issuecomment-290939569 and https://github.com/airbnb/enzyme/pull/898#issuecomment-294085827.

The Enzyme team recommends to call the `handlerFunction` directly for now.
It works well but exposes to test case an internal behavior of component that isn't a best practice.

The final solution will come when they add the `invoke` function to their API.
A PR already exists but it stills as a WIP: https://github.com/airbnb/enzyme/pull/1856

marcelomorgado:
> @marcelomorgado In the [new SDK version we published](https://github.com/tasitlabs/TasitSDK/issues/145), I think the function in question isn't async anymore, right?
Right

> Then again, I could see account creation getting slower again and changing to async depending on how we handle [tasitlabs/TasitSDK#24](https://github.com/tasitlabs/TasitSDK/issues/24)
:+1: 

Also, related to: https://github.com/tasitlabs/tasit/issues/42

Issue #31, opened 2018-12-18
Remove jest.preprocessor.js when the bug with jest on React Native v0.57 is resolved


pcowgill:
https://github.com/tasitlabs/tasit/issues/12#issuecomment-444184544

PR #48, opened 2018-12-19
WIP: enzyme click bug

marcelomorgado:
Refs: https://github.com/tasitlabs/tasit/issues/12#issuecomment-448300733

:warning: It isn't a branch to be merged. :warning: 

I've looked deeply about this issue and Enzyme team will not make "simulate" as an async function. Reasons: [here](https://github.com/airbnb/enzyme/pull/898#issuecomment-294085427), [here](https://github.com/airbnb/enzyme/pull/867#issuecomment-290939569) and [here](https://github.com/airbnb/enzyme/pull/898#issuecomment-294085827).

Since calling a component function isn't ideal because the test shouldn't be aware of component internal behavior, I've [found](https://github.com/airbnb/enzyme/pull/898) a trick to turn around that:

```
await wrapper.find("Button").find({ title: "Aync" }).prop("onPress")();
```
They are working on new function called [invoke](https://github.com/airbnb/enzyme/pull/1856) that will do the same as above:

```
await wrapper.find("Button").find({ title: "Aync" }).invoke("onPress");
await wrapper.find("Button").find({ title: "Aync" }).invoke("press");
```


pcowgill:
@marcelomorgado This is all great info. Thanks for looking into this! If we publish a new version of the SDK with the sync version of `Account.create()`, this problem will go away for now, right? Either way it's good to know our options for when we inevitably will need to use a button with an async function for onPress.

marcelomorgado:
> If we publish a new version of the SDK with the sync version of `Account.create()`, this problem will go away for now, right? 

Yes, It should work.

pcowgill:
@marcelomorgado If I recall correctly, we're keeping this open until we publish a new version of the SDK and then potentially merging it. Is that right?

marcelomorgado:
No, not really,

This PR isn't to be merged. It's for doc only. I was going to share this on Enzyme/React-native issues but it wasn't necessary.

There are three issues related with tests until now:

1) 
`simulate('click')` on component that has async handler function. It'll be solved with `invoke()` but this issue stills open: https://github.com/airbnb/enzyme/pull/1856
For now, we are calling async function directly from test case code.
Should I create an issue to track that?

2) 
The issue related with Expo SDK is that: https://github.com/tasitlabs/tasit/issues/12#issuecomment-448298933 (Snapshot)
But it stills occur because react-native used on SDK 32.0.0 is v0.57.1 (September)
The fix was introduced on react-native v.0.57.8 (December).

> Expo SDK v32.0.0 is based once again on [React Native 0.57 “September 2018”,](https://github.com/facebook/react-native/releases/tag/v0.57.1) the same major version as SDK 31.

Ref: https://blog.expo.io/expo-sdk-v32-0-0-is-now-available-6b78f92a6c52
Should I create an issue to track that?

3) 
`jest.preprocessor.js` file
We already have an issue: https://github.com/tasitlabs/tasit/issues/31

After creation of remain issues I think we could close this PR without merge.

marcelomorgado:
I've updated the issue https://github.com/tasitlabs/tasit/issues/43 and created the issue https://github.com/tasitlabs/tasit/issues/65.
I think that we could close this PR (without merge).

pcowgill:
Thanks for the status update. That all makes sense. Thanks!

Issue #51, opened 2019-01-07
Upgrade demo app to Expo SDK 32


pcowgill:

pcowgill:
I'm closing this since it was resolved in PR #57

PR #57, opened 2019-01-14
Upgrade expo SDK

marcelomorgado:
Refs #51


marcelomorgado:
Yes, we aren't using any deprecated imports/classes.

PR #67, opened 2019-01-19
Style the buttons

### LINKED ISSUE #38: Style the buttons
Make them look like buttons with a colored fill as opposed to hyperlinks

img_2911

marcelomorgado:
Closes #38 

## Note

I've found this (still beta) project: https://github.com/react-native-training/react-native-elements
It seems like a `material-ui` for `react-native`. Do you think that we can use this project (or similar) to handle with styling/cross-platform issues?

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -17,6 +17,7 @@
     "prop-types": "^15.6.2",
     "react": "16.5.0",
     "react-native": "https://github.com/expo/react-native/archive/sdk-32.0.0.tar.gz",
+    "react-native-button": "^2.3.0",
     "react-native-responsive-dimensions": "^2.0.1",
     "react-navigation": "^3.0.9",
     "tasit-sdk": "0.0.2"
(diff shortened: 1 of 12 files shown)


pcowgill:
I’m open to using it! At least in the demo app if not in all apps. And we should draw inspiration from particular components in it at a minimum.

Also one other option to consider linked here https://tasitlabs.slack.com/archives/CEZ8NCSTX/p1546626201000700

demo/package.json
```diff
@@ -17,6 +17,7 @@
     "prop-types": "^15.6.2",
     "react": "16.5.0",
     "react-native": "https://github.com/expo/react-native/archive/sdk-32.0.0.tar.gz",
+    "react-native-button": "^2.3.0",
```
pcowgill: For a component this small that may not be maintained it’s also tempting to incorporate the code directly in the codebase. But I like using it as a package here, at least for now

Issue #59, opened 2019-01-14
Add contract interaction to the demo app


pcowgill:
Using the features from the tasit-action package in the SDK

marcelomorgado:
Is this issue can be closed?

pcowgill:
@marcelomorgado Yep, I think so! I'll do that now. Thanks.

PR #79, opened 2019-02-05
Decentraland: Sell orders listing

marcelomorgado:
### Issue link

Refs: #59 

### Types of changes

### Notes

- Which is the better place to load config? HomeScreen.js, App.js or another place?

- I've used a babel config to remove all `console.*` from code. The app is faster without those.
**Refs**: https://docs.expo.io/versions/latest/react-native/performance/#using-consolelog-statements
That config is recommended for production since `console.*` can be useful on development. But, since `ethers.js` prints a lot of `warns` it's making the development process extremely slow (e.g.: Listing screen without `console.*` is taking about 2s, with `console.*` it's taking >10s!).
**Question**: Since our SDK is also prints warnings. Should we silent only `ethers.js`?

- I've changed `ListView` component to `FlatList` because performance too.
**Refs**: https://docs.expo.io/versions/latest/react-native/performance/#listview-initial-rendering-is-too-slow-or

- Should I rename the word `Land` to `SellOrder` on components names?

- How to get usd mana price? Static value or fetch from a market rate api?

- I'm not sure about the better place to keep contract addresses and objects, should we have a global assignment for that kind of objects?

- And about integration tests? Should we write some? I'm missing a test script since I've tested listing fetch on my mobile.

Thanks!


pcowgill:
Oh oops, I just approved part of the diff and accidentally approved the whole PR

pcowgill:
> * Which is the better place to load config? HomeScreen.js, App.js or another place?

App.js sounds good for now. Until that file gets too big, and then we can split it out.

> * I've used a babel config to remove all `console.*` from code. The app is faster without those.
>   **Refs**: https://docs.expo.io/versions/latest/react-native/performance/#using-consolelog-statements
>   That config is recommended for production since `console.*` can be useful on development. But, since `ethers.js` prints a lot of `warns` it's making the development process extremely slow (e.g.: Listing screen without `console.*` is taking about 2s, with `console.*` it's taking >10s!).
>   **Question**: Since our SDK is also prints warnings. Should we silent only `ethers.js`?

That's worth passing on as feedback to ethers.js, potentially.

Adding that Babel config is a great move.

Wow, that difference is dramatic. Yes, I think silencing ethers.js is the move in the short run. Or maybe not silencing it, but only passing through higher severity than `console.warn`

pcowgill:
> * I've changed `ListView` component to `FlatList` because performance too.
>   **Refs**: https://docs.expo.io/versions/latest/react-native/performance/#listview-initial-rendering-is-too-slow-or

Great, thanks. And thanks for explaining this in the notes.

> 
> * Should I rename the word `Land` to `SellOrder` on components names?

Hmm, I'd say it depends. If the things we're showing is mostly just the land as opposed to details about the order, then no. But if the component's level of abstraction is that of a piece of land plus details about the sale, then yes.

pcowgill:
> * How to get usd mana price? Static value or fetch from a market rate api?

Good question. I'm inclined to punt on this for now and remove USD from the views until `0.2.0`. But put a TODO in the code to add it back.

pcowgill:
> * I'm not sure about the better place to keep contract addresses and objects, should we have a global assignment for that kind of objects?

Can you include some pseudocode to clarify what you mean? My initial reaction is that yes, we should have a global assignment for this.

> * And about integration tests? Should we write some? I'm missing a test script since I've tested listing fetch on my mobile.

You mean like clicking with Enzyme? Or something involving interacting with ganache?

I think we should add integration tests soon-ish, but I don't think it's the best use of time until `0.2.0` or `0.3.0`. What do you think?

pcowgill:
@marcelomorgado FYI there's a merge conflict after merging other PRs. Can you please take a look when you get a chance? Thanks!

pcowgill:
@marcelomorgado Why not get the ABI from `tasit-contracts`?

marcelomorgado:
> > * I'm not sure about the better place to keep contract addresses and objects, should we have a global assignment for that kind of objects?
> 
> Can you include some pseudocode to clarify what you mean? My initial reaction is that yes, we should have a global assignment for this.
> 
Yes, sure:

```javascript
import { abi as estateABI } from "../abi/EstateRegistry.json";
import { abi as markplaceABI } from "../abi/Marketplace.json";
const { Contract } = Action;

const estateAddress = "0x41b598a2c618b59b74540ac3afffb32f7971b37a";
const marketplaceAddress = "0x07c0e972064e5c05f7b3596d81de1afd35457eae";
...
const estateContract = new Contract(estateAddress, estateABI);
const marketplaceContract = new Contract(marketplaceAddress, markplaceABI);
```
As we'll call these contracts from others components, I've thought to put them on an isolated class. For now they are living inside of `ListLandsScreen.js`. If so, which is the best practice, pattern for that?

> > * And about integration tests? Should we write some? I'm missing a test script since I've tested listing fetch on my mobile.
> 
> You mean like clicking with Enzyme? Or something involving interacting with ganache?
> 
> I think we should add integration tests soon-ish, but I don't think it's the best use of time until `0.2.0` or `0.3.0`. What do you think?

Please, tell me if I'm wrong but, I think that the unit tests from React apps, is to be only for test if rendering is working well, mocks and fakes should be used instead of a real blockchain/api/http call.

demo/components/presentational/LandRow.js
```diff
@@ -7,14 +7,22 @@ import {
...
-    <TouchableHighlight onPress={props.onPress}>
-      <View style={styles.row}>
-        <Land land={props.land} />
-      </View>
-    </TouchableHighlight>
-  );
+// Note: Changing to PureComponent for performance boost
+// It is possible to still using function component with React.memo HoC
+// See more:
+// https://reactjs.org/docs/react-api.html#reactpurecomponent
```
pcowgill: I think in React 16.7 there's a way to get the performance boost while keeping it written as a function. I'll try to dig up what I'm thinking of here. I'm pretty sure I'm thinking of something other than React.memo.

(I forget which React version React Native is using and which version of React Native Expo is using though)

marcelomorgado:
> > * I've changed `ListView` component to `FlatList` because performance too.
> >   **Refs**: https://docs.expo.io/versions/latest/react-native/performance/#listview-initial-rendering-is-too-slow-or
> 
> Great, thanks. And thanks for explaining this in the notes.
> 
> > * Should I rename the word `Land` to `SellOrder` on components names?
> 
> Hmm, I'd say it depends. If the things we're showing is mostly just the land as opposed to details about the order, then no. But if the component's level of abstraction is that of a piece of land plus details about the sale, then yes.

Hmm, got it. Said that, the `ListLandsScreen` is listing "estates sell orders" and the `LandClaimScreen`  and `Land` actually show price (a sell order info).

demo/config/default.js
```diff
@@ -0,0 +1,16 @@
+const development = {
```
pcowgill: I wonder whether we want to pick `default` or `development` for both the filename and the object?

demo/screens/HomeScreen.js
```diff
@@ -1,8 +1,15 @@
 import React from "react";
 import Home from "@presentational/Home";
 import { NavigationEvents } from "react-navigation";
+import { Action } from "tasit-sdk";
+import tasitSdkConfig from "../config/default.js";
 
 export default class HomeScreen extends React.Component {
+  componentDidMount() {
```
pcowgill: Let's make sure this screen shows something nice-looking if the connection to the blockchain can't be estabilished.

demo/screens/ListLandsScreen.js
```diff
@@ -1,60 +1,72 @@
...
-    img: require("../assets/images/lands/land4.png"),
-    priceMana: 150000,
-    priceUsd: 7500,
-  },
-];
-
-const rowHasChanged = (r1, r2) => r1.id !== r2.id;
-
-const ds = new ListView.DataSource({ rowHasChanged });
+const estateAddress = "0x41b598a2c618b59b74540ac3afffb32f7971b37a";
```
pcowgill: These can move to a constants file (like you mentioned in the PR notes)

demo/screens/ListLandsScreen.js
```diff
@@ -1,60 +1,72 @@
...
+    const rows = await this.getSellOrders();
+    this.setState({ rows });
+  }
+
+  async getSellOrders() {
+    const orders = [];
+    const estateContract = new Contract(estateAddress, estateABI);
+    const totalSupply = await estateContract.totalSupply();
+
+    for (let estateId = 1; estateId <= Number(totalSupply); estateId++)
```
pcowgill: Should we add some logic to only do this if the total supply is small?

demo/screens/ListLandsScreen.js
```diff
@@ -1,60 +1,72 @@
...
+      orders.push(this.getSellOrder(estateId));
```
pcowgill: My preferred style is to split this out into two lines

pcowgill:
```suggestion
      const order = this.getSellOrder(estateId);
      orders.push(order);
```

demo/screens/ListLandsScreen.js
```diff
@@ -1,60 +1,72 @@
...
+
+    return await Promise.all(orders);
+  }
+
+  async getSellOrder(estateId) {
+    const estateContract = new Contract(estateAddress, estateABI);
+    const marketplaceContract = new Contract(marketplaceAddress, markplaceABI);
```
pcowgill: Let's do this outside of this function

demo/screens/ListLandsScreen.js
```diff
@@ -1,60 +1,72 @@
...
+
+  async getSellOrders() {
+    const orders = [];
+    const estateContract = new Contract(estateAddress, estateABI);
+    const totalSupply = await estateContract.totalSupply();
+
+    for (let estateId = 1; estateId <= Number(totalSupply); estateId++)
+      orders.push(this.getSellOrder(estateId));
+
+    return await Promise.all(orders);
```
pcowgill: Are we assuming here that most estates will have a sell order?

demo/screens/ListLandsScreen.js
```diff
@@ -1,60 +1,72 @@
...
+  }
+
+  async getSellOrder(estateId) {
+    const estateContract = new Contract(estateAddress, estateABI);
```
pcowgill:
```suggestion
    // TODO: Switch to new DecentralandEstate() once the SDK includes that
    const estateContract = new Contract(estateAddress, estateABI);
```

demo/screens/ListLandsScreen.js
```diff
@@ -1,60 +1,72 @@
...
+    const marketplaceContract = new Contract(marketplaceAddress, markplaceABI);
+
+    const estateName = await estateContract.getMetadata(estateId);
+    const [
+      orderId,
```
pcowgill: Let's check for these being undefined before moving on in the function

demo/screens/ListLandsScreen.js
```diff
@@ -1,60 +1,72 @@
...
+      seller,
+      price,
+      expiresAt,
+    ] = await marketplaceContract.auctionByAssetId(estateId);
+
+    const priceMana = Number(price.toString()) / 1e18;
+    const manaUsdRate = 30;
```
pcowgill: A variable name that makes it clearer which is being divided by which would be useful. Like `manaPerUsd`

demo/screens/ListLandsScreen.js
```diff
@@ -1,60 +1,72 @@
...
+    const priceUsd = priceMana * manaUsdRate;
+
+    return {
+      id: estateId,
+      name: estateName,
+      priceMana,
+      priceUsd,
+      img: `https://api.decentraland.org/v1/estates/${estateId}/map.png`,
```
pcowgill: Build this in a temp var before the return?

demo/screens/ListLandsScreen.js
marcelomorgado: For the current local blockchain data, yes.

demo/screens/ListLandsScreen.js
marcelomorgado: I think that this function should be rewrite when we move to testnet. I've added a comment about that.

demo/screens/ListLandsScreen.js
marcelomorgado: Related to: https://github.com/tasitlabs/tasit/pull/79#discussion_r254306497

demo/components/presentational/LandRow.js
marcelomorgado: I've not found another way to use function components rather than React.memo.
We are using `react` 16.5.0
Expo 32.0.0 is using `react-native` v0.57.1 that uses `react` 16.5.0.

demo/config/default.js
marcelomorgado: Done. I think that maybe is desirable have more than one file inside `config/` (e.g.: development.js, production.js) and the app will only look for `default.js`. Make sense?

demo/screens/ListLandsScreen.js
marcelomorgado: Done.

demo/screens/ListLandsScreen.js
marcelomorgado: Done.

demo/screens/ListLandsScreen.js
marcelomorgado: Done.

demo/screens/ListLandsScreen.js
marcelomorgado: Done.

demo/components/presentational/LandRow.js
pcowgill: Cool, thanks for the info! I'll resolve this for now and let you know if I figure out what I have in mind. Thanks!

demo/config/default.js
pcowgill: Sure, that works for me! Let's just not put any of our API keys in production.js. We can have those injected during the build process in CI.

pcowgill: Resolving this

demo/screens/ListLandsScreen.js
pcowgill: Thanks! Resolving.

demo/screens/ListLandsScreen.js
pcowgill: Okay, perfect. Thanks! Resolving.

demo/screens/ListLandsScreen.js
pcowgill: Thanks! Resolving.

demo/screens/ListLandsScreen.js
pcowgill: Yep, we can resolve it based on what you said in the other thread.

demo/screens/ListLandsScreen.js
pcowgill: Thanks! Resolving.

demo/screens/ListLandsScreen.js
pcowgill: Thanks! Resolving.

demo/screens/ListLandsScreen.js
pcowgill: Thanks! Resolving.

demo/screens/HomeScreen.js
pcowgill: @marcelomorgado But we can just keep this in mind for the future. At most with a comment in the code.

demo/constants/ContractsABIs.js
```diff
@@ -0,0 +1,7 @@
+import { abi as estateABI } from "../abi/EstateRegistry.json";
```
pcowgill: I forget, did you have thoughts on getting the ABIs from over in the SDK in the future?

pcowgill: I suppose maybe that's not the best demo of how an external user would use it, though.

pcowgill: Mentioning this tool in case it's useful to think about, since it's kind of on this topic https://github.com/JoinColony/trufflepig

demo/screens/ListLandsScreen.js
```diff
@@ -1,60 +1,80 @@
...
+    rows: [],
   };
 
-  renderRow = land => {
+  async componentDidMount() {
+    const rows = await this.getSellOrders();
+    this.setState({ rows });
+  }
+
+  // Note: This function is assuing that:
```
pcowgill:
```suggestion
  // Note: This function is assuming that:
```

demo/screens/ListLandsScreen.js
```diff
@@ -19,39 +21,45 @@ export default class ListLandsScreen extends React.Component {
...
+    ] = await this.marketplaceContract.auctionByAssetId(estateId);
+
+    const hasOrder = parseInt(orderId, 16) !== 0;
+    if (!hasOrder) throw Error(`Estate (id:${estateId}) has no sell order.`);
 
     const priceMana = Number(price.toString()) / 1e18;
-    const manaUsdRate = 30;
-    const priceUsd = priceMana * manaUsdRate;
+    const manaPerUsd = 30;
+    const priceUsd = priceMana * manaPerUsd;
```
pcowgill: Hmm wait, wouldn't the price in USD be a lower number than the price in mana? So should we divide by 30?

demo/screens/HomeScreen.js
marcelomorgado: OK, should I create an issue for that too?

demo/screens/ListLandsScreen.js
marcelomorgado: Oh yes, my fault! Done.

demo/constants/ContractsABIs.js
marcelomorgado: > I forget, did you have thoughts on getting the ABIs from over in the SDK in the future?

Yes, maybe we can expose ABIs when Decentraland package will be available;

> I suppose maybe that's not the best demo of how an external user would use it, though.

Using the Contract class with their own ABI isn't the user will do something like that?
Maybe expose some ABIs (20,721,etc)  as constants?

> Mentioning this tool in case it's useful to think about, since it's kind of on this topic
https://github.com/JoinColony/trufflepig

Maybe we have to think on both scenarios: 
1) When user under develop process using the SDK (local blockchain / abi from `truffle`/`trufflepig`)
2) When SDK is using to interact with a pre-existing mainnet (without ganache-cli / not deployed by user)

Brainstorm Note: When I was writing item 2, I've thought that we can even suggest to our users development stages when interact with 3rd-party contracts. As we are doing with Decentraland:
- Local deployment from same code and test as contract owner
- Local network fork from testnet (We will do that and test suite should be changed because we'll no longer been owner of contracts);
- Testnet network (Should be a smooth move from last step)
- Local network fork from mainnet;
- Mainnet network.
Assuming that the 3rd-party keeps both testnet/mainnet as Decentraland does.

ABI Note: What do you think about `ethers.js` human readable ABI? Should we allows users instante contracts using that?

demo/screens/ListLandsScreen.js
marcelomorgado: Done, sorry.

demo/screens/ListLandsScreen.js
pcowgill: Thanks! Resolving!

marcelomorgado:
Open threads:
- Rename components (Land vs. SellOrder);
- Silent console.log & open `ethers.js` issue;
- USD Removal.
- Integration test bootstrap.
Could I handle with these on next PR to merge that sooner?

demo/screens/ListLandsScreen.js
pcowgill: Thanks! Resolving!

pcowgill:
@marcelomorgado Yep, let’s handle those in a future PR

demo/constants/ContractsABIs.js
pcowgill: On my mobile device, but my initial reaction is that that all makes sense and sounds good! Should we extract this discussion to an issue?

marcelomorgado: https://github.com/tasitlabs/tasit/issues/82

pcowgill: Thanks! Resolving

demo/screens/HomeScreen.js
pcowgill: Sure. Thanks. Resolving this.

pcowgill: Or wait, I shouldn't resolve it until we create the issue. But you can resolve this.

pcowgill:
> Open threads:
> 
>     * Rename components (Land vs. SellOrder);
> 
>     * Silent console.log & open `ethers.js` issue;
> 
>     * USD Removal.
> 
>     * Integration test bootstrap.
>       Could I handle with these on next PR to merge that sooner?

@marcelomorgado I forget, do we have an issue for these open threads yet?

marcelomorgado:
Not yet, I'll create, I got it.

demo/screens/HomeScreen.js
marcelomorgado: :+1:

marcelomorgado: https://github.com/tasitlabs/tasit/issues/83. Resolving this.

marcelomorgado:
https://github.com/tasitlabs/tasit/issues/84

Issue #84, opened 2019-02-06
PR 79 Open threads


marcelomorgado:
Refs: https://github.com/tasitlabs/tasit/pull/79#issuecomment-461084152

* Rename components (Land vs. SellOrder): https://github.com/tasitlabs/tasit/pull/79#issuecomment-460845469
* Silent console.log & open `ethers.js` issue: https://github.com/tasitlabs/tasit/pull/79#issuecomment-460835222
* USD Removal: https://github.com/tasitlabs/tasit/pull/79#issuecomment-460836068

* Integration test bootstrap. **Note: I've just saw that you said to do that after 0.1.0.**

marcelomorgado:
`ethers.js` feedback issue: https://github.com/ethers-io/ethers.js/issues/416

Issue #60, opened 2019-01-15
Solve test warning messages


marcelomorgado:
The test script is outputting some `prop type` warning messages.

Refs: https://github.com/tasitlabs/tasit/pull/58#issuecomment-454212853

marcelomorgado:
I think that issue can be closed.

PR #85, opened 2019-02-06
PR #79 open threads

marcelomorgado:
Closes: https://github.com/tasitlabs/tasit/issues/84, https://github.com/tasitlabs/tasit/issues/60

Note: I'm assuming that "integration test boostrap" should be done after v0.1.0. Am I right?


pcowgill:
> Note: I'm assuming that "integration test boostrap" should be done after v0.1.0. Am I right?

Yep, I think so. So should we split that out into a separate issue?

demo/babel.config.js
```diff
@@ -2,6 +2,9 @@ module.exports = function(api) {
   api.cache(true);
   return {
     presets: ["babel-preset-expo"],
-    plugins: ["transform-remove-console"],
+
+    // Note: console.* should be removed from production app because performance issues
+    // See more: https://docs.expo.io/versions/latest/react-native/performance/#using-consolelog-statements
+    //plugins: ["transform-remove-console"],
```
pcowgill: Why comment it out?

marcelomorgado:
> Yep, I think so. So should we split that out into a separate issue?
Yes, I'll create an issue.

Note: Last CI failed, I've just committed the fix but seems that CircleCI is facing instability right now.

demo/components/presentational/EstateForSale.js
```diff
@@ -0,0 +1,27 @@
...
+import PropTypes from "prop-types";
+import {
+  responsiveWidth,
+  responsiveHeight,
+} from "react-native-responsive-dimensions";
+
+export default function EstateForSale({ estate }) {
+  return (
+    <React.Fragment>
+      <Image style={styles.landImage} source={{ uri: estate.img }} />
```
pcowgill: Maybe split out the with `const { img}  = estate` first?

demo/components/presentational/EstateForSale.test.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { shallow } from "enzyme";
+import EstateForSale from "./EstateForSale";
+
+describe("EstateForSale", () => {
+  it("renders the component", async () => {
+    const estate = {
+      id: -1,
```
pcowgill: Maybe a random number would be better in case we ever add validation on the integers being positive?

marcelomorgado:
Integration test issue: https://github.com/tasitlabs/tasit/issues/86

demo/components/presentational/EstateForSale.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import { Image, StyleSheet, Text } from "react-native";
+import PropTypes from "prop-types";
+import {
+  responsiveWidth,
+  responsiveHeight,
+} from "react-native-responsive-dimensions";
+
+export default function EstateForSale({ estate }) {
```
pcowgill: What about it matters that it's for sale here? Maybe just `Estate`?

demo/components/presentational/SellOrder.js
```diff
@@ -0,0 +1,28 @@
...
+
+export default function SellOrder({ sellOrder }) {
+  const { asset: estate, priceMana } = sellOrder;
+
+  // Note: Conversion to USD will be implemented on v0.2.0
+  // <Text>{priceMana} MANA (~${sellOrder.priceUsd})</Text>
+  return (
+    <View style={styles.landContainer}>
+      <EstateForSale estate={estate} />
+      <Text>{priceMana} MANA</Text>
```
pcowgill: I wonder whether we want to style the number and the word mana with different fonts, fontsizes, fontweights, etc.

demo/components/presentational/SellOrder.test.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { shallow } from "enzyme";
+import SellOrder from "./SellOrder";
+
+describe("SellOrder", () => {
+  it("renders the component", async () => {
+    const sellOrder = {
+      id: -1,
+      priceMana: 0,
+      priceUsd: 0,
```
pcowgill: Maybe `priceUSD` since USD is already an acronym

demo/components/presentational/SellOrdersList.test.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+import { shallow } from "enzyme";
+import SellOrdersList from "./SellOrdersList";
+
+describe("SellOrdersList", () => {
+  it("renders the component", async () => {
+    expect(
+      shallow(<SellOrdersList sellOrders={[]} renderItem={() => {}} />)
```
pcowgill: Might read more easily with the function that does nothing assigned to a var before this line

demo/package.json
```diff
@@ -51,7 +51,8 @@
       "\\.js$": "<rootDir>/jest.preprocessor.js"
     },
     "setupFiles": [
-      "<rootDir>/jest.setup.js"
+      "<rootDir>/jest.setup.js",
+      "jest-prop-type-error"
```
pcowgill: What does this do?

demo/screens/ListSellOrdersScreen.js
```diff
@@ -49,32 +50,42 @@ export default class ListLandsScreen extends React.Component {
...
-      id: estateId,
-      name: estateName,
+      id: orderId,
       priceMana,
       priceUsd,
-      img: imgUrl,
-      orderId,
       seller,
       expiresAt,
+      asset: {
```
pcowgill: Is it asset so that it could work with a different ERC721 NFT as well? Or just land also? Or will asset here always be an estate in the future?

demo/screens/ListSellOrdersScreen.js
```diff
@@ -49,32 +50,42 @@ export default class ListLandsScreen extends React.Component {
...
-    const { item: land } = row;
+  renderItem = ({ item: sellOrder }) => {
     const handlePress = () =>
-      this.props.navigation.navigate("LandClaimScreen", { land });
-
-    return <LandRow id={land.id} land={land} onPress={handlePress} />;
+      this.props.navigation.navigate("SellOrderScreen", { sellOrder });
+    return (
+      <SellOrdersListItem
+        id={sellOrder.id}
```
pcowgill: Why pass in id inside the `sellOrder` object and separately?

demo/babel.config.js
marcelomorgado: Because it's silent expo warnings and useful debug console.log too.
I think that with `ethers.js` loglevel last change the performance issue was addressed.

demo/components/presentational/EstateForSale.test.js
marcelomorgado: A random number will fail the snapshot test. I'll change for a positive one. Ok?

demo/components/presentational/EstateForSale.js
marcelomorgado: Yeah, I agree.

demo/package.json
marcelomorgado: Without that, proptype errors are ignored by jest.
Refs: https://github.com/tasitlabs/tasit/issues/60 (I think that change solves this issue).
Now, when test case faces a proptype error, it will fail.

pcowgill: Cool. I'll resolve this thread.

demo/components/presentational/EstateForSale.test.js
pcowgill: Sorry, I meant an "arbitrary" number, not "random". Sounds good. Thanks!

demo/babel.config.js
pcowgill: But I thought it only happened in production? So you would still see useful debug console.log while developing locally, right?

demo/screens/ListSellOrdersScreen.js
marcelomorgado: Marketplace contract accept any ERC721 as asset, on Decentraland context, it can be parcel or estate, I'm not sure if it's the best approach, I've thought on both solutions:
Using `asset` field with some type field to indicate which asset it is, 
or
Using `estate` / `parcel` fields and check which isn't `undefined` to know which asset it is.
What do you think?

demo/screens/ListSellOrdersScreen.js
marcelomorgado: I've seen that way on the `FlatList` doc and I've thought that was mandatory but it isn't. I've just removed that.

demo/components/presentational/EstateForSale.js
marcelomorgado: Done.

demo/components/presentational/EstateForSale.test.js
marcelomorgado: Done.

demo/components/presentational/EstateForSale.js
marcelomorgado: Done.

demo/components/presentational/SellOrder.test.js
marcelomorgado: Done.

demo/components/presentational/SellOrdersList.test.js
marcelomorgado: Done.

demo/screens/ListSellOrdersScreen.js
marcelomorgado: Done.

demo/components/presentational/SellOrder.js
marcelomorgado: Yes, probably. Sorry but styling isn't mt strength...

marcelomorgado: I'm couldn't think on any suggestion here.

demo/screens/ListSellOrdersScreen.js
pcowgill: I think I prefer asset field with an enumerable list of strings for type.

demo/screens/ListSellOrdersScreen.js
pcowgill: Thanks! Resolving.

demo/components/presentational/EstateForSale.js
pcowgill: Thanks! Resolving.

demo/components/presentational/EstateForSale.test.js
pcowgill: Thanks! Resolving.

demo/components/presentational/EstateForSale.js
pcowgill: Thanks! Resolving.

demo/components/presentational/SellOrder.test.js
pcowgill: Thanks! Resolving.

demo/components/presentational/SellOrdersList.test.js
pcowgill: Thanks! Resolving.

demo/babel.config.js
marcelomorgado: Right

demo/components/presentational/SellOrder.js
pcowgill: Okay! I'll split this out into a separate issue.

pcowgill: https://github.com/tasitlabs/tasit/issues/87

Resolving here

demo/babel.config.js
marcelomorgado: The "definitive" solution here is use the plugin only on prod build, I've tryied for a while to work with babel/expo `env` params but has some workarounds to do so I thought better keep that way for know and improve build scripts later.

pcowgill: So then maybe we should uncomment this?

pcowgill: I made my comment above just before you made your last one.

Okay, I assumed the plugin worked only in prod by default or something, so I was wrong about that.

Makes sense and sounds good! We can always work on this again later when it comes up again.

pcowgill: Resolving this

demo/screens/ListSellOrdersScreen.js
marcelomorgado: I've created a comment about that.

pcowgill: Thanks. Resolving.

demo/components/presentational/SellOrdersList.test.js
```diff
@@ -4,8 +4,12 @@ import SellOrdersList from "./SellOrdersList";
 
 describe("SellOrdersList", () => {
   it("renders the component", async () => {
+    const renderItem = () => {};
```
pcowgill: In the future this function could maybe use a clearer name. Not going to gate merging.

marcelomorgado: Is `itemRenderer` a good name?

pcowgill: I like that better, yes. I think the best might be ____Renderer where you say what the item is instead of using a generic word like `item`

Issue #62, opened 2019-01-18
Use contract function via the Tasit SDK to buy land using the Decentraland marketplace contracts


pcowgill:

pcowgill:
@marcelomorgado This issue would probably move the project forward a long way in terms of end-to-end completeness, because it would force us to [publish the latest version of the Tasit SDK](https://github.com/tasitlabs/TasitSDK/issues/145) and show how it could be used from within a demo app.

For that reason, I think it's a strong next choice for an issue to work on. But since I have the publishing permissions for the project on npm still, it probably would make sense to work on together towards the end of your work day.

pcowgill:
﻿﻿﻿This is about to be merged, so that will unblock this PR https://github.com/tasitlabs/TasitSDK/issues/145

pcowgill:
@marcelomorgado Think this should be the next major issue to work on?

pcowgill:
If not, maybe just the "updating the demo app to use the latest SDK version" part could be split out into a separate issue that we tackle first

marcelomorgado:
I'm thinking to finish the issue: https://github.com/tasitlabs/TasitSDK/issues/91#issuecomment-457782599
With test cases that we'll use on the app and then go into on this one. Make sense?

pcowgill:
@marcelomorgado Sure, that sounds good!

marcelomorgado:
Is this issue can be closed?

pcowgill:
@marcelomorgado Yep! Closing it now. Thanks! Great work.

PR #100, opened 2019-02-12
Using Decentraland contracts objects (ABI files removal)

### LINKED ISSUE #94: Remove .tasitid.eth from the pick username screen
The user shouldn't need to wonder what ENS is when doing this

marcelomorgado:
### Issue link

Refs #62 
Closes #94 

### Types of changes

Other (please describe below if you select "Other")

### Notes

Other: Fix "Lands" grammar


demo/screens/BuyLandScreen.js
```diff
@@ -4,19 +4,18 @@ import { removeLandForSale } from "../actions";
...
+const { Estate } = ERC721;
+const { Decentraland } = Marketplace;
 import { approveManaSpending, manaFaucetTo } from "./helpers";
 
 export class BuyLandScreen extends React.Component {
-  // TODO: Switch to new DecentralandEstate() once the SDK includes that
-  estateContract = new Contract(estateAddress, estateABI);
-  marketplaceContract = new Contract(marketplaceAddress, marketplaceABI);
+  estateContract = new Estate(ESTATE_ADDRESS);
+  marketplaceContract = new Decentraland(MARKETPLACE_ADDRESS);
```
pcowgill: Hmmm. Now that I'm seeing this it feels awkward that it's called Decentraland. Maybe as a short-term change we can import it `as` another name?

marcelomorgado: Yes,
`const { Decentraland: Marketplace } = Marketplace` sounds good?

pcowgill: Works for me!

marcelomorgado: Done.
I've used `const { Decentraland: DecentralandMarketplace } = Marketplace`
because `Marketplace` var is already defined.
Or you prefer?
`const { Decentraland: Marketplace } = Action.Marketplace`

marcelomorgado:
Last commit closes: https://github.com/tasitlabs/tasit/issues/94

demo/screens/BuyLandScreen.js
pcowgill: The option you went with works for me. Tough choice, haha. Resolving this.

Issue #98, opened 2019-02-12
Change components folder structure


marcelomorgado:
As @mseijas recommended on https://github.com/tasitlabs/tasit/issues/32#issuecomment-462585784
Is this change have to be done only for presentational components?

pcowgill:
Hmm, good question. Let's just do it with presentational components to start. We can always do it for the container components / screens in a later PR if we find that that's helpful too.

Issue #32, opened 2018-12-18
Use Storybook to show our presentational components


pcowgill:
This will help us to identify which components are general enough that they belong in a `tasit-components` dir at the top level of this repo (that is, outside of the demo app). That way they can be used in other apps in this repo or used by the `tasit-cli` when generating an app scaffold for developers using the `tasit-sdk`.

It will also be nice to link to the storybook from the developers page on the website for the project.

https://storybook.js.org/
https://storybook.js.org/basics/guide-react-native/

pcowgill:
@nirmalb9 This issue in particular would be a good one for someone interested in front-end / mobile

mseijas:
@pcowgill I'm giving this a shot. Thoughts on changing the folder structure a bit so that we have one folder per component?

This way we can keep all of these together, say:

``` 
  /Button
     /index.js (exports Button)
     /Button.js
     /Button.test.js
     /Button.story.js
```

pcowgill:
@mseijas Thanks!! That works for me. @marcelomorgado any objections?

marcelomorgado:
No, no objections. I've created an issue for that: https://github.com/tasitlabs/tasit/issues/98
Is this change have to be done only for presentational components?

pcowgill:
Hmm, good question. Let's just do it with presentational components to start. We can always do it for the container components / screens in a later PR if we find that that's helpful too.

pcowgill:
I responded about presentational components vs. container components over in the other issue,

PR #104, opened 2019-02-12
Implement Storybook

mseijas:
### Auto-close the issue?

Closes https://github.com/tasitlabs/tasit/issues/32
Closes https://github.com/tasitlabs/tasit/issues/98

### Types of changes

### Notes

Note that the entry point of the app has been changed from `App.js` to `index.js`

`index.js` determines if the app should be loaded normally, or if storybook should be loaded instead. This is defined by the value passed to the `SHOW_STORYBOOK` boolean.

Additionally, there's a prestorybook hook that will run before the storybook server is executed, and that automatically generates `/storybook/storyLoader.js` and includes all the files inside `/components/` that match `**/*.story.js`


marcelomorgado:
Thanks @mseijas!

We've refactored some components recently:

SellOrder was renamed to LandForSale
SellOrderClaim was renamed to BuyLand
SellOrderList was renamed to LandForSaleList
SellOrderListItem was renamed to LandForSaleListItem

Please, tell me if you need some help on conflicts because the code was changed/renamed too.

A question: Is possible to use storyboard from web to see the components instead of mobile?

mseijas:
@marcelomorgado Hey so I tried rebasing with the new changes and it's been kind of a mess - there are so many changes to apply that I'm not confident that things are looking right after rebasing. I'm thinking it may be simpler to just start again on top of the newly merged changes to reduce the risk of overwriting or breaking anything. 

I'm thinking of maybe splitting it into different PRs to keep track of the changes more easily. Maybe one with folder reorganization and another one with storybook implementation. What are your thoughts on this?

marcelomorgado:
> @marcelomorgado Hey so I tried rebasing with the new changes and it's been kind of a mess - there are so many changes to apply that I'm not confident that things are looking right after rebasing. I'm thinking it may be simpler to just start again on top of the newly merged changes to reduce the risk of overwriting or breaking anything.
> 
> I'm thinking of maybe splitting it into different PRs to keep track of the changes more easily. Maybe one with folder reorganization and another one with storybook implementation. What are your thoughts on this?

Hi,
Yeah, work from the updated `develop` seems a good idea and split PRs too.
Thanks!

pcowgill:
@mseijas Yep, that sounds good. Sorry the timing for this PR ended up making it so complicated! I’ll close this one then.

pcowgill:
@mseijas Feel free to message me on Twitter if you're about to work on this again, because I'll want to make sure there aren't any PRs in progress that would make it complicated for you a second time. Thanks!

mseijas:
@pcowgill will do! this week is a bit crazy for me, so I’ll probably circle back to this early next week

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

pcowgill:
@mseijas Sounds good. Thanks!!

Issue #101, opened 2019-02-12
Share Decentraland app using Expo


pcowgill:
@marcelomorgado We can try this even before releasing the app through TestFlight

marcelomorgado:
https://expo.io/@marcelo.morgado/tasit-demo

pcowgill:
Awesome, thanks!

Until this is done: https://expo.canny.io/feature-requests/p/better-team-collaboration-tools-or-workflow

Let’s create a user named tasit to publish the app under and share the password (outside of GitHuB)

pcowgill:
@marcelomorgado Can you share permissions to view the app you published with the the expo user `pcowgill`? Thanks! I don’t think I can view it in the app until that is done.

marcelomorgado:
Weird... The app is public, can you see the qrcode?
Has no option to share permissions. Are you sure that exists such an option?
Could you create the `tasit` user (since I have no tasit email account)?

pcowgill:
Maybe it’s that we didn’t build it for iOS?

Yep, I’ll create the user

pcowgill:
![aada11d6-64c4-4d06-8b87-b6fc75d48916](https://user-images.githubusercontent.com/2731712/52719885-25489d00-2f6c-11e9-932c-f1c3bed328ea.png)

marcelomorgado:
> Maybe it’s that we didn’t build it for iOS?

Yes, you are right. I'll build for ios now.

> Yep, I’ll create the user

:+1:

pcowgill:
I opened the app using the button since I’m on iOS.

pcowgill:
![0bdbbe4d-a263-475d-bd9a-d4d696ce5f71](https://user-images.githubusercontent.com/2731712/52721059-73f73680-2f6e-11e9-81c3-e47e223658e2.png)

marcelomorgado:
Yeah, I haven't built the iOS version yet, the Expo is asking for an Apple ID. How to proceed?

pcowgill:
Ah okay, you can pause there for now and work on something else? I’ll set up an Apple id

marcelomorgado:
Sure. :+1:

pcowgill:
Sharing the demo app using Expo can only be done on Android with the new iOS rules. So I think we can mark this as complete.

Although maybe we should re-publish the demo app under the `tasit` expo username before closing this.

pcowgill:
[...]

PR #122, opened 2019-03-11
Supporting iOS build and publish

marcelomorgado:
### Issue link

#101 

### Types of changes

### DIFF
--- a/demo/app.json
+++ b/demo/app.json
@@ -19,10 +19,11 @@
     },
     "assetBundlePatterns": ["**/*"],
     "ios": {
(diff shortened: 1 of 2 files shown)


pcowgill:
@marcelomorgado Thanks! What should we do with https://github.com/tasitlabs/tasit/pull/105?

marcelomorgado:
> @marcelomorgado Thanks! What should we do with #105?

Oh, I forget that PR, this one replaces PR#105. I think we should merge that and close PR#105 without merging.

PR #124, opened 2019-03-11
Removal of the jest.preprocessor.js file

### LINKED ISSUE #31: Remove jest.preprocessor.js when the bug with jest on React Native v0.57 is resolved

marcelomorgado:
### Issue link

#31 

### Auto-close the issue?

Closes #31 

### Types of changes


marcelomorgado:
I've added the `--maxWorkers=2` option because without that some test were got completed on Circle CI. That option solved the problem after search about that problem on one of the Circle CI forum threads.

> ...Circle CI provides 2 vCPUs so --maxWorkers=2 might give better performance than --runInBand

Refs: https://discuss.circleci.com/t/babel-7-jest-results-in-timeout/25383/4

pcowgill:
@marcelomorgado We already updated the React Native version outside of this PR, right?

marcelomorgado:
> @marcelomorgado We already updated the React Native version outside of this PR, right?

Yes

PR #149, opened 2019-03-18
Decentraland app flow using ropsten testnet

marcelomorgado:
### Issue link

#99 
### Auto-close the issue?

Closes #99 

### Types of changes

New feature (non-breaking change that adds functionality)

Meta (a change that affects future changes to this repo)

Technical debt (a code change that doesn't fix a bug or add a feature but makes something clearer for devs)

Issue #88, opened 2019-02-07
Use cross-platform UI components


marcelomorgado:
@pcowgill said recently:
> I recently heard from one engineer that he prefers react-native-elements to NativeBase

Some modules:
https://github.com/react-native-training/react-native-elements
https://github.com/IjzerenHein/react-native-magic-move
https://github.com/GeekyAnts/NativeBase

Note: I'm not sure which milestone that belongs.

pcowgill:
0.2.0 seems like a reasonable milestone for this.

pcowgill:
Cross-platform UI components:

https://github.com/react-native-training/react-native-elements
https://github.com/GeekyAnts/NativeBase

Motion / animations:
https://github.com/IjzerenHein/react-native-magic-move
https://popmotion.io/pose/learn/native-get-started/

pcowgill:
I think we can extract the motion / animations into a different issue, since it's pretty unrelated

marcelomorgado:
Extracted to: https://github.com/tasitlabs/tasit/issues/215

pcowgill:
@marcelomorgado I think we can close this issue

marcelomorgado:
Yes, I agree. Closing this.

Issue #151, opened 2019-03-19
Using events instead of `waitForNonceToUpdate`


marcelomorgado:
UX should take advantage of events from contracts and actions.

marcelomorgado:
@pcowgill commented:

> I'm wondering whether the await style still makes sense for higher-level functions that wrap Ethereum interactions. I think it doesn't but we'll want to set best practices for the right pattern when using this style to not end up in nested "callback hell"
>
> The callback style of listening for success with a subscription/callback makes the most sense when used along with optimistic updates, which we aren't doing here yet.

Refs: https://github.com/tasitlabs/tasit/pull/163#discussion_r267405172

pcowgill:
@marcelomorgado I think this will follow nicely after the SDK updates. Using conf’s is worth considering in combination with or as an alternative to using events

pcowgill:
(These SDK updates https://github.com/tasitlabs/TasitSDK/issues/219)

pcowgill:
> Yes, we are almost there, I think we have only a few bugs to fix before to do this.
> [tasitlabs/TasitSDK#368](https://github.com/tasitlabs/TasitSDK/issues/368), [tasitlabs/TasitSDK#367](https://github.com/tasitlabs/TasitSDK/issues/367), [tasitlabs/TasitSDK#366](https://github.com/tasitlabs/TasitSDK/issues/366)

Great! And yep, agreed that those are gating. Thanks for sharing the specific issues.

marcelomorgado:
Updating:

~~https://github.com/tasitlabs/TasitSDK/issues/368~~ Closed
https://github.com/tasitlabs/TasitSDK/issues/367 fixed on PR https://github.com/tasitlabs/TasitSDK/pull/383

marcelomorgado:
I think that issue https://github.com/tasitlabs/TasitSDK/issues/384 should be solved as well.

pcowgill:
> Updating:
> 
> ~[tasitlabs/TasitSDK#368](https://github.com/tasitlabs/TasitSDK/issues/368)~ Closed
> [tasitlabs/TasitSDK#367](https://github.com/tasitlabs/TasitSDK/issues/367) fixed on PR [tasitlabs/TasitSDK#383](https://github.com/tasitlabs/TasitSDK/pull/383)

Great!

pcowgill:
[...]

PR #163, opened 2019-03-20
UX improvements

### LINKED ISSUE #159: UX: Use a proper component to show messages to user
For now, we are using `YellowBox` warnings and that should be improved.

Note: Errors from blockchain events and from thrown errors should use this solution.

### LINKED ISSUE #160: UX improvement: Don't allow user to buy until their account is prepared
Immediately after the account creation the user is redirected to the `BuyLandScreen` but the fund and approve transactions are waiting. The `buy` button should be disallowed until all that setup is finished.

marcelomorgado:
### Issue link

#160 
#159 
#88
https://github.com/tasitlabs/tasit/pull/149#discussion_r266668330

### Auto-close the issue?

Closes #160 
Closes #159 

### Types of changes

### Notes

I've focused on UX over UI/beauty.


decentraland/.prettierignore
```diff
@@ -0,0 +1 @@
+native-base-theme/
```
pcowgill: What's the deal with this dir?

pcowgill:
@marcelomorgado What motivated the choice of `NativeBase` over `react-native-elements`? https://github.com/tasitlabs/tasit/issues/88#issue-407857669

Either is a sensible choice, so just curious for your thought process. Thanks!

decentraland/screens/BuyLandScreen.js
```diff
@@ -36,11 +37,20 @@ export class BuyLandScreen extends React.Component {
...
+    const { navigation, accountInfo, removeLandForSale } = props;
+    const { account } = accountInfo;
+    const { type } = landForSale;
+
+    if (type !== ESTATE && type !== PARCEL) showError(`Unknown asset.`);
+
+    const typeDescription = type == ESTATE ? "Estate" : "Parcel";
+
     const onSuccess = () => {
+      showInfo(`${typeDescription} bought successfully.`);
```
pcowgill: Nice

decentraland/screens/EthereumSignUpScreen.js
```diff
@@ -13,31 +13,41 @@ import {
...
 
-      await approveManaSpending(account);
-      showInfo(`Account created and funded!`);
+      const onFundSuccess = async () => {
+        const onApprovalSuccess = () => {
+          setSetupInProgress(false);
+          showInfo(`Account created and funded!`);
+        };
+
+        approveManaSpending(account, onApprovalSuccess);
```
pcowgill: I'm wondering whether the await style still makes sense for higher-level functions that wrap Ethereum interactions. I think it doesn't but we'll want to set best practices for the right pattern when using this style to not end up in nested "callback hell"

The callback style of listening for success with a subscription/callback makes the most sense when used along with optimistic updates, which we aren't doing here yet.

Larger topic - probably worth a discussion outside of this thread. This is already a major improvement as is! Nice work!

pcowgill: The thing that made me think about this was the two separate steps of funding the account and then approving  mana spending. I think we may want to have a single callback that awaits the success of both operations with a promise.all style?

marcelomorgado:
> NativeBase

`NativeBase` seems that have more complete range of components. The toast component is an example of that, `react-native-elements` hasn't such a component.
But it isn't a strong opinion since I have no experience on none of them.

decentraland/screens/EthereumSignUpScreen.js
```diff
@@ -13,31 +13,41 @@ import {
...
 
   _onSignUp = () => {
-    showInfo(`Creating and funding account...`);
-
-    // Should run async but isn't when calling Account.create() or createFromPrivateKey()
-    this._onboarding();
+    const { setSetupInProgress } = this.props;
+    setSetupInProgress(true);
+    // Note: A trick to force `_onboarding()` function to running async
+    (async () => {})().then(() => {
```
pcowgill: Interesting. Nice hack! We can continue to ponder any alternate strategies but this will serve us very well in the meantime.

decentraland/screens/helpers.js
```diff
@@ -50,20 +52,21 @@ export const getContracts = () => {
...
+};
+
+export const approveManaSpending = async (fromAccount, onSuccess) => {
   const contracts = getContracts();
   const { manaContract, marketplaceContract } = contracts;
   const toAddress = marketplaceContract.getAddress();
   manaContract.setWallet(fromAccount);
   const action = manaContract.approve(toAddress, TEN);
   await action.waitForNonceToUpdate();
+  onSuccess();
```
pcowgill: If we're going to use the `onSuccess` callback, we could probably get rid of the `await` of `waitForNonceToUpdate` and use our listening style?

decentraland/screens/helpers.js
```diff
@@ -85,14 +88,16 @@ export const fundAccount = async accountAddress => {
...
+const showToast = msg => Toast.show({ text: msg, duration: 3000 });
 export const showFatalError = msg => console.error(msg);
-export const showError = msg => console.warn(`ERROR: ${msg}`);
-export const showWarn = msg => console.warn(`WARN: ${msg}`);
-export const showInfo = msg => console.warn(`INFO: ${msg}`);
-export const showSuccess = msg => console.warn(`SUCCESS: ${msg}`);
+export const showError = msg => showToast(`ERROR: ${msg}`);
+export const showWarn = msg => showToast(`WARN: ${msg}`);
+export const showInfo = msg => showToast(`INFO: ${msg}`);
+export const showSuccess = msg => showToast(`SUCCESS: ${msg}`);
```
pcowgill: Thoughts on `buttonText: 'Okay'` for dismissing in some cases?

pcowgill:
> `NativeBase` seems that have more complete range of components. The toast component is an example of that, `react-native-elements` hasn't such a component.
> But it isn't a strong opinion since I have no experience on none of them.

Makes sense - yeah, I just saw that there is nothing like a toast in React Native Elements. Happy with the choice in that case! Choosing one or the other isn't necessarily mutually exclusive

pcowgill:
When I first reviewed this PR I'm not sure I saw that all the NativeBase components were included in the diff

decentraland/.prettierignore
marcelomorgado: Hmmm, since those files were auto-generated I'm not sure if we should change them, should we?

decentraland/screens/EthereumSignUpScreen.js
marcelomorgado: A) Yeah, I think we should refactor the blockchain interactions on the app to establish a pattern / good practices soon (I've extracted your comment to the issue https://github.com/tasitlabs/tasit/issues/151)
B) Yes, make sense. I'll do that.

decentraland/screens/helpers.js
marcelomorgado: Sure!

decentraland/screens/EthereumSignUpScreen.js
marcelomorgado: B) Done.

decentraland/screens/helpers.js
marcelomorgado: Hmmm.. on my last change we aren't using callback no more:
```javascript
      const fund = fundAccount(accountAddress);
      const approve = approveManaSpending(account);
      await Promise.all([fund, approve]);
      showInfo(`Account created and funded!`);
```

decentraland/screens/helpers.js
marcelomorgado: I've just added this option to all cases, there is a case that shouldn't have that?

decentraland/screens/helpers.js
pcowgill: Ah ok. Well we can continue the discussion thread in the other thread and likely move it off GitHub since it's a larger topic. That approach looks good for now.

decentraland/screens/EthereumSignUpScreen.js
pcowgill: Great, thanks. We can continue the rest of the discussion over in that issue.

pcowgill: Resolving this

decentraland/screens/helpers.js
pcowgill: Or yeah, like we said in the other thread, we can move the discussion to this issue https://github.com/tasitlabs/tasit/issues/151

Resolving this

decentraland/screens/EthereumSignUpScreen.js
pcowgill: Resolving this

decentraland/.prettierignore
pcowgill: It's a little weird they're included in the repo as opposed to being available in `node_modules`. I agree that we shouldn't change the auto-generated files unless we plan on essentially "forking" them and maintaining them outside of the dep - and I don't think we want to do that in the short-term or medium-term.

decentraland/screens/BuyLandScreen.js
pcowgill: Resolving this

decentraland/screens/helpers.js
pcowgill: Thanks. I think the only time we wouldn't want it to be dismissible is if the toast were going to be around until something was complete. But I don't think a toast is the right UI element for that type of situation. So I think having it in all cases is fine.

decentraland/.prettierignore
pcowgill: I think this is the main question gating merging

marcelomorgado: Without running the generation script, the module doesn't work. Maybe we can keep that way until we need to make any changes.

pcowgill: What's the generation script? Is it mentioned here? https://docs.nativebase.io/docs/GetStarted.html

marcelomorgado: Sorry @pcowgill , I've just made a double check about what I've said and the module **does** work without this folder. The script is this: `node node_modules/native-base/ejectTheme.js` (refs: https://docs.nativebase.io/Customize.html#theme-and-variables-headref) and is used to customized the default components. I've run that because I've faced a problem with fonts setup and I've thought that ejection was part of the solution.

marcelomorgado: I've just committed the removal.

pcowgill: Ah ok, I figured I'd check since it seemed a little odd. It makes sense that customization led you in that direction, but good to hear it wasn't strictly necessary!

decentraland/screens/helpers.js
pcowgill: Resolving this

PR #194, opened 2019-03-26
No assets for sale message

### LINKED ISSUE #187: Add a message on screen when have no assets for sale
When we have no assets for sale, the screen stays blank without feedback for the user, seeming that is buggy.

marcelomorgado:
### Issue link

#187 
### Auto-close the issue?

Closes #187 

### Types of changes

### DIFF
--- a/decentraland/components/presentational/LandForSaleList.test.js
+++ b/decentraland/components/presentational/LandForSaleList.test.js
@@ -6,11 +6,13 @@ describe("LandForSaleList", () => {
   it("renders the component", async () => {
     const landForSaleRenderer = () => {};
     const landForSaleList = [];
+    const loadingInProgress = true;
     expect(
       shallow(
         <LandForSaleList
           landForSaleList={landForSaleList}
           renderItem={landForSaleRenderer}
+          loadingInProgress={loadingInProgress}
         />
       )
     ).toMatchSnapshot();
--- a/decentraland/components/presentational/LandForSaleList.test.js
+++ b/decentraland/components/presentational/LandForSaleList.test.js
@@ -3,18 +3,35 @@ import { shallow } from "enzyme";
 import LandForSaleList from "./LandForSaleList";
 
 describe("LandForSaleList", () => {
-  it("renders the component", async () => {
-    const landForSaleRenderer = () => {};
-    const landForSaleList = [];
-    const loadingInProgress = true;
-    expect(
-      shallow(
-        <LandForSaleList
-          landForSaleList={landForSaleList}
-          renderItem={landForSaleRenderer}
-          loadingInProgress={loadingInProgress}
-        />
-      )
-    ).toMatchSnapshot();
+  describe("renders the component", () => {
+    it("when list is empty and the loading is in progress", async () => {
+      const landForSaleRenderer = () => {};
+      const landForSaleList = [];
+      const loadingInProgress = true;
+      expect(
+        shallow(
+          <LandForSaleList
+            landForSaleList={landForSaleList}
+            renderItem={landForSaleRenderer}
+            loadingInProgress={loadingInProgress}
+          />
+        )
+      ).toMatchSnapshot();
+    });
+
+    it("after loaded an empty list", async () => {
+      const landForSaleRenderer = () => {};
+      const landForSaleList = [];
+      const loadingInProgress = false;
+      expect(
+        shallow(
+          <LandForSaleList
+            landForSaleList={landForSaleList}
+            renderItem={landForSaleRenderer}
+            loadingInProgress={loadingInProgress}
+          />
+        )
+      ).toMatchSnapshot();
+    });
   });
 });
--- a/decentraland/components/presentational/LandForSaleList.test.js
+++ b/decentraland/components/presentational/LandForSaleList.test.js
@@ -19,7 +19,7 @@ describe("LandForSaleList", () => {
       ).toMatchSnapshot();
     });
 
-    it("after loaded an empty list", async () => {
+    it("after having loaded an empty list", async () => {
       const landForSaleRenderer = () => {};
       const landForSaleList = [];
       const loadingInProgress = false;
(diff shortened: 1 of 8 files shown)


decentraland/components/presentational/LandForSaleList.test.js
```diff
@@ -6,11 +6,13 @@ describe("LandForSaleList", () => {
   it("renders the component", async () => {
     const landForSaleRenderer = () => {};
     const landForSaleList = [];
+    const loadingInProgress = true;
     expect(
       shallow(
         <LandForSaleList
           landForSaleList={landForSaleList}
           renderItem={landForSaleRenderer}
+          loadingInProgress={loadingInProgress}
```
pcowgill: We probably want a separate test in this file that does a `loadingInProgress` `false` render

decentraland/screens/ListLandForSaleScreen.js
```diff
@@ -147,29 +156,34 @@ export class ListLandForSaleScreen extends React.Component {
...
         renderItem={this._renderItem}
+        loadingInProgress={loadingInProgress}
       />
     );
   }
 }
 
 ListLandForSaleScreen.propTypes = {
-  landForSaleList: PropTypes.array.isRequired,
+  assetsForSale: PropTypes.object.isRequired,
```
pcowgill: I wonder if we should have kept what's now called `assetsForSale` as an array and had a separate prop not inside the object for the loading boolean

decentraland/components/presentational/LandForSaleList.test.js
```diff
@@ -6,11 +6,13 @@ describe("LandForSaleList", () => {
   it("renders the component", async () => {
```
pcowgill: It's odd the snapshot for this file didn't change in this PR

decentraland/components/presentational/LandForSaleList.test.js
marcelomorgado: Done.

decentraland/screens/ListLandForSaleScreen.js
marcelomorgado: Hmmm,
Both `assetsForSale` and `accountInfo` has their loading flags as an inside prop. I don't have a strong opinion if we should keep that way or extract the flags to the same level.
I think that keep that way has a pro to keep cohesion but it adds more complexity since it adds more nested objects.

decentraland/components/presentational/LandForSaleList.test.js
marcelomorgado: It is the desirable behavior, in my first try (only check the array) the message was shown before the first asset load. Because of that, I've introduced the `loadingInProgress` flag and the "no assets" message only be showed when loading as completed and the list is empty.
In resume, empty list + inprogress = true has the same behavior as before (blank page).

decentraland/components/presentational/LandForSaleList.test.js
pcowgill: Thanks! Resolving

decentraland/screens/ListLandForSaleScreen.js
pcowgill: Okay - that makes sense and definitely won't gate merging this PR. This will be an interesting topic to ponder together going forward. I think having them as separate loading states might force us to consider whether we have a reasonable amount of loading states. We might even move all loading states at some point into a dedicated loading state section of state.

pcowgill: Should we extract this to an issue or just keep it in mind?

decentraland/components/presentational/LandForSaleList.test.js
pcowgill: Ah ok - I guess what surprised me was that this test was doing an expectation on the blank page as opposed to the full page. If it was doing an expectation on a blank page before, then this diff makes sense.

pcowgill: It sounds like in a future PR we'll want a third placeholder rendered view for this screen that says "Loading"

decentraland/components/presentational/LandForSaleList.test.js
```diff
@@ -3,16 +3,35 @@ import { shallow } from "enzyme";
...
+          <LandForSaleList
+            landForSaleList={landForSaleList}
+            renderItem={landForSaleRenderer}
+            loadingInProgress={loadingInProgress}
+          />
+        )
+      ).toMatchSnapshot();
+    });
+
+    it("after loaded an empty list", async () => {
```
pcowgill:
```suggestion
    it("after having loaded an empty list", async () => {
```

decentraland/components/presentational/__snapshots__/LandForSaleList.test.js.snap
```diff
@@ -1,5 +1,11 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`LandForSaleList renders the component after having loaded an empty list 1`] = `
+<LargeText>
+  There are no assets for sale right now.
+</LargeText>
+`;
+
 exports[`LandForSaleList renders the component after loaded an empty list 1`] = `
```
pcowgill: @marcelomorgado Hmm, I guess I should delete this manually since this is the one with the old name?

pcowgill: I ended up deleting it. Interesting to know that jest doesn't do that automatically

marcelomorgado: I'll do that

marcelomorgado: Oh, I just saw your last message :+1:

decentraland/screens/ListLandForSaleScreen.js
marcelomorgado: I think that an issue will be better: https://github.com/tasitlabs/tasit/issues/197

decentraland/components/presentational/LandForSaleList.test.js
marcelomorgado: > It sounds like in a future PR we'll want a third placeholder rendered view for this screen that says "Loading"

Yes, for sure. The UX needs to be improved with more feedback to the user on that way.

decentraland/screens/ListLandForSaleScreen.js
pcowgill: Okay, sounds good. Resolving this.

PR #205, opened 2019-03-27
Disable button during account creation

### LINKED ISSUE #178: Gray out the buy button while it's in "waiting for account" mode
Screenshot 2019-03-20 15 13 17

marcelomorgado:
### Issue link
#178 

### Auto-close the issue?

Closes #178 

### Types of changes

Other (please describe below if you select "Other")

### Notes

UI improvement

### DIFF
--- a/decentraland/components/presentational/Button.js
+++ b/decentraland/components/presentational/Button.js
@@ -16,8 +16,10 @@ const BUTTON_HEIGHT = 1.5;
 // https://github.com/tasitlabs/tasit/issues/204
 export default function Button(props) {
   let { title, onPress, disabled } = props;
-  if (!disabled) disabled = false;
-  if (!onPress) onPress = () => {};
+  if (disabled === undefined) disabled = false;
+
+  const onPressHandler = disabled ? () => {} : onPress;
+
   title = title.toUpperCase();
 
   return (
@@ -26,7 +28,7 @@ export default function Button(props) {
         disabled ? containerStyles.disabled : containerStyles.enabled
       }
       style={disabled ? buttonStyles.disabled : buttonStyles.enabled}
-      onPress={onPress}
+      onPress={onPressHandler}
       activeOpacity={1}
     >
       {title}
@@ -36,7 +38,7 @@ export default function Button(props) {
 
 Button.propTypes = {
   title: PropTypes.string.isRequired,
-  onPress: PropTypes.func,
+  onPress: PropTypes.func.isRequired,
   disabled: PropTypes.bool,
 };
 
--- a/decentraland/components/presentational/Button.js
+++ b/decentraland/components/presentational/Button.js
@@ -23,9 +23,9 @@ export default function Button(props) {
   return (
     <RNButton
       containerStyle={
-        disabled ? disabledStyles.container : enabledStyles.container
+        disabled ? containerStyles.disabled : containerStyles.enabled
       }
-      style={disabled ? disabledStyles.button : enabledStyles.button}
+      style={disabled ? buttonStyles.disabled : buttonStyles.enabled}
       onPress={onPress}
       activeOpacity={1}
     >
@@ -40,28 +40,28 @@ Button.propTypes = {
   disabled: PropTypes.bool,
 };
 
-const enabledStyles = StyleSheet.create({
-  button: {
-    fontSize: responsiveFontSize(FONT_SIZE),
-    color: Colors.buttonColor,
-  },
-  container: {
+const containerStyles = StyleSheet.create({
+  enabled: {
     padding: responsiveHeight(BUTTON_HEIGHT),
     overflow: "hidden",
     borderRadius: 3,
     backgroundColor: Colors.buttonBackground,
   },
-});
-
-const disabledStyles = StyleSheet.create({
-  button: {
-    fontSize: responsiveFontSize(FONT_SIZE),
-    color: Colors.disabledButtonColor,
-  },
-  container: {
+  disabled: {
     padding: responsiveHeight(BUTTON_HEIGHT),
     overflow: "hidden",
     borderRadius: 3,
     backgroundColor: Colors.disabledButtonBackground,
+  },
+});
+
+const buttonStyles = StyleSheet.create({
+  disabled: {
+    fontSize: responsiveFontSize(FONT_SIZE),
+    color: Colors.disabledButtonColor,
+  },
+  enabled: {
+    fontSize: responsiveFontSize(FONT_SIZE),
+    color: Colors.buttonColor,
   },
 });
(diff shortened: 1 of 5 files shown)


pcowgill:
“action” -> “account”?

marcelomorgado:
> “action” -> “account”?

Issue name adjusted

decentraland/components/presentational/Button.js
```diff
@@ -8,34 +8,56 @@ import {
   responsiveFontSize,
 } from "react-native-responsive-dimensions";
 
+// TODO: Change to NativeBase button component
+// https://github.com/tasitlabs/tasit/issues/204
 export default function Button(props) {
+  let { title, onPress, disabled } = props;
+  if (!disabled) disabled = false;
+  if (!onPress) onPress = () => {};
```
pcowgill: Are we sure we want this to be an optional prop that gets defaulted?

decentraland/components/presentational/Button.js
```diff
@@ -8,34 +8,56 @@ import {
...
+  title = title.toUpperCase();
+
   return (
     <RNButton
-      containerStyle={styles.container}
-      style={styles.button}
-      onPress={props.onPress}
+      containerStyle={
+        disabled ? disabledStyles.container : enabledStyles.container
```
pcowgill: I might have done `container.disabledStyle` personally so I could more quickly see what was different from the enabled version of the container. But this is a perfectly good choice too!

decentraland/components/presentational/Button.js
```diff
@@ -8,34 +8,56 @@ import {
...
-
     overflow: "hidden",
     borderRadius: 3,
     backgroundColor: Colors.buttonBackground,
   },
 });
+
+const disabledStyles = StyleSheet.create({
+  button: {
+    fontSize: responsiveFontSize(1.8),
```
pcowgill: Since 1.8 is used twice, make it a `const` above?

decentraland/components/presentational/__snapshots__/Button.test.js.snap
```diff
@@ -19,6 +19,29 @@ exports[`Button renders the component 1`] = `
...
+  ENABLED BUTTON
+</Button>
+`;
+
+exports[`Button renders the component disabled button 1`] = `
+<Button
+  activeOpacity={1}
+  containerStyle={
+    Object {
+      "backgroundColor": "gray",
```
pcowgill: It's nice when reviewing PRs to have the snapshots in here to see that indeed it is gray in this case

decentraland/components/presentational/Button.js
marcelomorgado: Which prop?

decentraland/components/presentational/Button.js
marcelomorgado: Yes, done.

decentraland/components/presentational/__snapshots__/Button.test.js.snap
marcelomorgado: Yeah :+1:

decentraland/components/presentational/Button.js
pcowgill: If I say "this", I usually mean the exact line I'm commenting on, which is always the lowest one in the preview on the GitHub UI.

So in this case, I mean the `onPress` one

If it's a broader concept, I try to make sure to comment on the lowest relevant line, so that any lines above you that need to see for context will show up.

decentraland/components/presentational/Button.js
pcowgill: Thanks! Resolving

decentraland/components/presentational/__snapshots__/Button.test.js.snap
pcowgill: Resolving this

decentraland/components/presentational/Button.js
pcowgill: Want to change it before merging or keep it the same?

decentraland/components/presentational/Button.js
marcelomorgado: Oh, ok. Thank you to clarify that! The UI is highlighting three lines. Got it now.
No strong opinion, doing another look maybe it isn't a good option. Maybe make more sense having `onPress` as optional if `disabled` is true. I've looked that it is possible since `Proptype` supports conditional requirement but I'm not sure if it worth increasing the complexity that way.
So, should I back `onPress` as a required prop?

decentraland/components/presentational/Button.js
marcelomorgado: Done. It's much better now!

pcowgill: Thanks! Resolving

decentraland/components/presentational/Button.js
pcowgill: I think the way to go is to make `onPress` required even for disabled buttons, and then internally in the button component have the function that passed the `onPress` prop into the handler for the button do an if on whether disabled is true or not and don't call the `onPress` function that was passed in if disabled is true.

pcowgill: That way the user can either pass in the real function that would be called if the button were enabled (which leaves us open for logic that just toggles the disabled state but keeps everything else the same) or pass in a mock function that doesn't do anything

pcowgill: I agree that conditional requirement of props is too much complexity for now (and maybe is always a code smell), so I like that the strategy above gets around that

pcowgill:
@marcelomorgado I think once this is merged it's worth rolling out a new version to testers

decentraland/components/presentational/Button.js
marcelomorgado: > I think the way to go is to make `onPress` required even for disabled buttons, and then internally in the button component have the function that passed the `onPress` prop into the handler for the button do an if on whether disabled is true or not and don't call the `onPress` function that was passed in if disabled is true.

Sorry @pcowgill but I'm not sure if I've got your message, please correct me if I'm wrong:
1. `onPress` as required
2. Use `onPress` prop on the button if `disabled` != true.

Is it?

pcowgill: > Sorry @pcowgill but I'm not sure if I've got your message, please correct me if I'm wrong:
>    1. onPress as required
>    2. Use onPress prop on the button if disabled != true.
>
> Is it?

Yep! For 2, I might word it as still use a function wrapping onPress on the button regardless, but only execute the onPress code within that function if `diabled` != `true`

marcelomorgado: Done.

pcowgill: Thanks! Resolving

Issue #42, opened 2018-12-18
Slow back after clicking continue on EthereumSignUpForm


pcowgill:
EthereumSignUpForm-> Continue (Account.create call) is delaying react-navigation. After a click, going back to the previous screen is laggy

pcowgill:
I'm moving this to `0.2.0` because I think we can assume the account is set up already and hard-code it in `0.1.0`

marcelomorgado:
Ohhh... I've just seen that is to work with an already set up account on `0.1.0`, I've worked on solving this issue. I'll continue "Buy a land" use case using a "from privateKey" Account. 
Sorry!

Anyway, I'll share what I've found until now:


## Report - React-Native performance (Account.create())

First of all, I’ve implemented a screen changing after SignUp screen to get UI feedback of createAccount() fired by onPress event of the button.

After that, all cache was cleaned ([reference](https://forums.expo.io/t/how-to-clear-the-react-native-packager/1352/14)):
  
```bash
~/../demo$ rm .expo -rf
~/../demo$ rm node_modules/ -rf
~/../demo$ npm cache clean --force
~/../demo$ npm i
```

Base scenario: After clicking on button (that call account creation) the screen changes after about 20 seconds. The code:

  
```javascript  
export default class EthereumSignUpForm extends React.Component {
  ...
  createAccount = () => {
    const wallet = Account.create();
    this.setState({ address: wallet.address });
  };

  // After the button clicked, the console.log is called after 20s
  onContinue = () => {
    console.log("button clicked");
    this.createAccount();
    this.props.afterSignUp();
  };

  render() {
    return (
      ...
      <Button title="Continue" onPress={() => this.onContinue()} />
      ....
    );
  }
}
```

**Attempt:** I’ve used Expo to build and generate Android APK file, the intent here was to eliminate debug network connection as the cause of issue.
**Result:** No changes.

**Attempt:** Removal of all `console.*` (uncomment babel …. )
**Result:** No changes.

**Attempt:** Changing `onPress` handler to `async`
**Result:** No changes.

**Attempt:** Comment `createAccount()` line
**Result:** Handle function called instantiatly.
**Conclusion:** The issue seems happens with `Account.create()` command, but why the `console.log` is called after 20s isn't clear.

**Attempt:** Assuming that the slow code is account creation one, the code was changed screen and run the account creation async. The one modifiction on code above is making `createAccount` function `async`
**Result:** No changes.
[...]

PR #207, opened 2019-03-27
Persist account

### LINKED ISSUE #190: Persist ephemeral account between app quitting and reopening

marcelomorgado:
### Issue link

#190 
### Auto-close the issue?

Closes #190 

### Types of changes

### Notes

Marked as `WIP` because it wasn't tested properly yet.

I'm not able to test close and reopen app since `Expo` rebuild app each time that it's opened. Is there a problem if I build/publish an `android` `expo` package without changing version, only to test it?


pcowgill:
> I'm not able to test close and reopen app since Expo rebuild app each time that it's opened. Is there a problem if I build/publish an android expo package without changing version, only to test it?

I think this is a perfect use case for [Expo release channels](https://docs.expo.io/versions/latest/distribution/release-channels/). We can publish a new version after still incrementing patch version, and any previous builds will get the latest js assets and we can test it that way. I think it's fine to increment version on published builds even when we don't end up needing to create a new .ipa or .apk build.

Note that a build would already need to have been made pointing to a non-prod release channel for this to work.

Publishing new js assets for Android to test it on a prod release channel would be bad in case it breaks the experience for testers

A new build is only necessary relatively rarely when we have OTA updates enabled (although I think for security/trust/auditability reasons when we roll the app out to a wider audience we need to disable OTA updates)

marcelomorgado:
> > I'm not able to test close and reopen app since Expo rebuild app each time that it's opened. Is there a problem if I build/publish an android expo package without changing version, only to test it?
> 
> I think this is a perfect use case for [Expo release channels](https://docs.expo.io/versions/latest/distribution/release-channels/). 

Nice feature!

> We can publish a new version after still incrementing patch version, and any previous builds will get the latest js assets and we can test it that way. 

:+1: 

> I think it's fine to increment version on published builds even when we don't end up needing to create a new .ipa or .apk build.

So, we'll increment version even for release-channels builds?

> Note that a build would already need to have been made pointing to a non-prod release channel for this to work.

OK, any release-channel naming convention?

> Publishing new js assets for Android to test it on a prod release channel would be bad in case it breaks the experience for testers

A question, even if I have done just an `expo publish:android` it'll update android users app somehow?

> A new build is only necessary relatively rarely when we have OTA updates enabled (although I think for security/trust/auditability reasons when we roll the app out to a wider audience we need to disable OTA updates)

:+1:

decentraland/screens/helpers.js
```diff
@@ -55,6 +62,40 @@ export const getContracts = () => {
...
+    await SecureStore.setItemAsync(key, value);
+  } catch (error) {
+    showError("Unable to save data on storage.");
+  }
+};
+
+const _retrieveData = async key => {
+  try {
+    const value = await SecureStore.getItemAsync(key);
+    return value;
```
pcowgill: `A promise that resolves to the previously stored value, or null if there is no entry for the given key. The promise will reject if an error occurred while retrieving the value.` Source: https://docs.expo.io/versions/latest/sdk/securestore/#returns-1

Should we check if value is `null` and log anything special, or is that outside of the scope of this function and something that should happen in "userland"?

decentraland/screens/helpers.js
```diff
@@ -55,6 +62,40 @@ export const getContracts = () => {
...
+  if (recoveredPrivateKey == null) {
+    // Note: The timeout for account creation is about ~20 secs.
+    // See more: https://github.com/tasitlabs/tasit/issues/42
+    const account = Account.create();
+    const { privateKey: createdPrivateKey } = account;
+    await _storeData(EPHEMERAL_ACCOUNT_PRIV_KEY, createdPrivateKey);
+    return account;
+  }
+
+  const account = createFromPrivateKey(recoveredPrivateKey);
```
pcowgill: I suppose we could also persist some things from the account object itself if we don't want to frequently re-derive it? I'll have to give that more thought.

decentraland/screens/helpers.js
marcelomorgado: Yeah, I think it's a "userland" matter since `null` can be an OK return.

pcowgill:
> So, we'll increment version even for release-channels builds?

Hmm, I'd say use our best judgement on a case-by-case basis. Certainly overkill to do this many times while working on a single feature

> OK, any release-channel naming convention?

Let's keep it simple to start.

Either `dev` (for development) or `stag` (for staging) would work for the purpose you described here.

And then for now we can keep using the default of `default` for the `prod` (production) build?

> A question, even if I have done just an expo publish:android it'll update android users app somehow?

Yes, but only while we have OTA updates turned on (most Expo users do, but we won't in the future). So be careful with that, please.

More info: https://docs.expo.io/versions/latest/guides/configuring-ota-updates/

**One last topic on OTA updates:**
If I recall correctly, iOS does it on next open of the app before the code runs, and Android starts loading next open of the app but doesn't run the new code until the following time the app is opened. OR vice versa - I'd need to check

decentraland/screens/helpers.js
pcowgill: Sounds good. Resolving this!

decentraland/screens/helpers.js
pcowgill: That won't gate merging this PR, though. I think it's out-of-scope for this PR.

marcelomorgado: Maybe `my assets` list? Refs: https://github.com/tasitlabs/tasit/issues/210

marcelomorgado:
> > So, we'll increment version even for release-channels builds?
> 
> Hmm, I'd say use our best judgement on a case-by-case basis. Certainly overkill to do this many times while working on a single feature

OK, so for example, to one branch test like this, seems better keeping the version. And update that when will test a 
group of changes before build to prod. Make sense?

> > OK, any release-channel naming convention?
> 
> Let's keep it simple to start.
> 
> Either `dev` (for development) or `stag` (for staging) would work for the purpose you described here.
> 
> And then for now we can keep using the default of `default` for the `prod` (production) build?

Worth create npm scripts for these?

> > A question, even if I have done just an expo publish:android it'll update android users app somehow?
> 
> Yes, but only while we have OTA updates turned on (most Expo users do, but we won't in the future). So be careful with that, please.
> 
> More info: https://docs.expo.io/versions/latest/guides/configuring-ota-updates/
> 
> **One last topic on OTA updates:**
> If I recall correctly, iOS does it on next open of the app before the code runs, and Android starts loading next open of the app but doesn't run the new code until the following time the app is opened. OR vice versa - I'd need to check

OK :+1:

decentraland/screens/helpers.js
pcowgill: That would be interesting to persist too. But I mean even persisting details of the account object itself that can be generated from the private key but might be slightly faster just to have ready.

Do you remember where that old issue was where you timed how long the `createFromPrivateKey` function takes to run?

pcowgill:
> OK, so for example, to one branch test like this, seems better keeping the version. And update that when will test a group of changes before build to prod. Make sense?

Sounds good!

> Worth create npm scripts for these?

Yep!

decentraland/screens/helpers.js
marcelomorgado: The issue is it: https://github.com/tasitlabs/tasit/issues/42
And I timed the `Account.create()` function, it's taking 20seg (I've counted again today).
I'll do that for `createFromPrivateKey` too.

decentraland/screens/helpers.js
```diff
@@ -55,6 +62,40 @@ export const getContracts = () => {
   return contracts;
 };
 
+// Note: the value should be a string
+const _storeData = async (key, value) => {
+  try {
+    await SecureStore.setItemAsync(key, value);
```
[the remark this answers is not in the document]
marcelomorgado: Yes, I'll do that.

decentraland/screens/helpers.js
pcowgill: Okay great - thanks!

decentraland/screens/helpers.js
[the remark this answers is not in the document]
marcelomorgado: Done.

decentraland/screens/helpers.js
marcelomorgado: `createFromPrivateKey` runs instantly

marcelomorgado:
I'll run a android `dev` build after your review on npm scripts.

decentraland/screens/helpers.js
pcowgill: Oh ok awesome, that's good to know

pcowgill: In that case, no need to persist the rest of the account object separately

pcowgill: Resolving this

decentraland/screens/helpers.js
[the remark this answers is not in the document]
pcowgill: Thanks! Resolving

decentraland/package.json
```diff
@@ -5,9 +5,13 @@
...
   "private": true,
   "scripts": {
-    "build:android": "npx expo build:android",
     "start": "FORCE_COLOR=1 node starter.js",
+    "build:android": "npx expo build:android",
     "build:ios": "npx expo build:ios",
+    "build:dev:android": "npx expo build:android --release-channel dev",
+    "build:dev:ios": "npx expo build:ios --release-channel dev",
+    "build:stag:android": "npx expo build:android --release-channel stag",
+    "build:stag:ios": "npx expo build:ios --release-channel stag",
```
pcowgill: I'm not sure if we'll use both the dev and stag channels just yet, but I suppose it can't hurt to have both of them configured now

pcowgill:
> I'll run an android `dev` build after your review on npm scripts.

Feel free to run the `dev` build now

decentraland/package.json
marcelomorgado: I've thought to create only the `dev` script, but I created `stag` too because I've thought that you could want it too.

pcowgill: Yep, makes sense to me. Resolving this!

marcelomorgado:
> > I'll run an android `dev` build after your review on npm scripts.
> 
> Feel free to run the `dev` build now

Done but it isn't working yet.

marcelomorgado:
**Notes**
- I was wrong, Expo handles store data, the previous code that wasn't working.
- The `App.js` seemed for me the best place to put that kind of 'resource loading' but I'm not sure if the way that I've stored the account (calling `redux` store directly) is an elegant solution.

pcowgill:
> I was wrong, Expo handles store data, the previous code that wasn't working.

@marcelomorgado Can you please rephrase this ^ ? Not sure I follow yet.

pcowgill:
>   The `App.js` seemed for me the best place to put that kind of 'resource loading' but I'm not sure if the way that I've stored the account (calling `redux` store directly) is an elegant solution.

`App.js` seems like a reasonable place to me too, at least for now. I'll take a look at the use of `redux` store

decentraland/App.js
```diff
@@ -6,6 +6,8 @@ import AppNavigator from "./AppNavigator";
 import { Provider } from "react-redux";
 import { createStore } from "redux";
 import decentralandApp from "./redux/reducers";
+import { setAccount } from "./redux/actions";
+import { recoverAccount } from "./screens/helpers";
```
pcowgill: Now that we're importing this in `App.js` it feels a little weird to get it from `screens/helpers`.

decentraland/screens/EthereumSignUpScreen.js
```diff
@@ -9,19 +9,15 @@ import {
...
 export class EthereumSignUpScreen extends React.Component {
   _onboarding = async () => {
     try {
       const { setAccount, setSetupInProgress } = this.props;
 
-      // Note: The timeout for account creation is about ~10 secs.
-      // See more: https://github.com/tasitlabs/tasit/issues/42
-      const account = Account.create();
-      setAccount(account);
+      const account = await createAccount();
```
pcowgill: Are we making sure that they only get to this step on the `EthereumSignUpScreen` if they don't already have an account?

marcelomorgado:
> > I was wrong, Expo handles store data, the previous code that wasn't working.
> 
> @marcelomorgado Can you please rephrase this ^ ? Not sure I follow yet.

I've said: 
> I'm not able to test close and reopen app since Expo rebuild app each time that it's opened.

But this's wrong, using `Expo` app I was able to reopen the app and it got the account from storage.

decentraland/App.js
```diff
@@ -44,6 +46,11 @@ Is the 'config/default.js' file correct?`;
     }
   }
 
+  async _loadAccount() {
+    const account = await recoverAccount();
+    if (account != null) store.dispatch(setAccount(account));
```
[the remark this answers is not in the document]
marcelomorgado: Hmmm, are you suggesting that we can always get the account from the storage? If yes, I think that will work but I'm not sure if it will be a good practice since the recommendation is having a "single source of truth".

decentraland/App.js
marcelomorgado: Should we move to a `@helpers` subpackage?

decentraland/screens/EthereumSignUpScreen.js
marcelomorgado: Yes, the Sign-up/Sign-in flow are called only if the account isn't setted:

**from BuyLandScreen:**
```javascript
  _onBuy = landForSale => {
    try {
      const { accountInfo } = this.props;
      const { account } = accountInfo;
      if (!account) this._setupAccount();
      // The _buy function assumes that the account is created, funded and allowed
      // The user will get back to this else case after going through an account setup flow
      else this._buy(landForSale);
    } catch (err) {
      showError(err);
    }
  };
```

pcowgill:
> But this's wrong, using `Expo` app I've reopened the app and it got the account from storage.

Oh okay, great!

decentraland/App.js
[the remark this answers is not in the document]
pcowgill: Yeah, I was. But I think your argument is a convincing one, so I'll resolve this.

decentraland/App.js
pcowgill: Something that we would publish to npm or just a dir with a package.json with a name so we can reference it like `@helpers`?

Are you picturing it would be outside of the decentraland app dir to share with other apps or inside it?

decentraland/screens/EthereumSignUpScreen.js
pcowgill: Ah right - okay, perfect. Thanks for explaining that! Resolving this.

decentraland/App.js
pcowgill: For the purposes of this PR, something as simple as moving the helpers file outside of the `screens` dir would work.

marcelomorgado: I meant just a dir with a package.json.
Ok, I've moved all functions from `screens/helpers.js` to `helpers.js`

marcelomorgado: Oh, Wait... I've found a bug here.

pcowgill: > Oh, Wait... I've found a bug here.

Okay, noted!

pcowgill: > I meant just a dir with a package.json.
> Ok, I've moved all functions from `screens/helpers.js` to `helpers.js`

Sounds good, thanks

marcelomorgado: I've merged both helpers, the `starter.js` script uses a helper function and because of that I installed `babel-node` since the `helpers` is using ES6 imports. But it's throwing error because `babel` isn't transpiling the `expo` `SecureStore` import. To fix that quickly I've copied the function from `helper` to `starter`.
I can extract the `checkBlockchain` function to another helper file and use it (to avoid copied code) but it'll add one more file to the root dir. I think that the best solution could be create the "package helpers dir" that will have a `index.js` and `startHelpers.js`, what do you think?

decentraland/helpers.js
```diff
@@ -1,9 +1,158 @@
...
+import {
+  Account,
+  Action,
+  ContractBasedAccount,
+  TasitContracts,
+} from "tasit-sdk";
 const { ConfigLoader } = Action;
-const { ProviderFactory } = require("tasit-action/dist/ProviderFactory");
-const tasitSdkConfig = require("./config/default.js");
+import ProviderFactory from "tasit-action/dist/ProviderFactory";
```
pcowgill: It's interesting that the switch to remove `require` and the switch to remove the `{ }` coincided. Is that expected? I'm sure you've tried it, so probably.

decentraland/starter.js
```diff
@@ -1,6 +1,22 @@
 /* eslint no-console: "off" */
-const { exec } = require("child_process");
-const { checkBlockchain } = require("./helpers.js");
+import { exec } from "child_process";
+
+// Note: Copied from helpers.js because import isn't working
+//import { checkBlockchain } from "./helpers";
```
pcowgill: Interesting

decentraland/App.js
pcowgill: That all makes sense. Your last suggestion sounds great! But that can come in a follow-up PR.

decentraland/helpers.js
marcelomorgado: Yes, it is. I've made this sometimes before:
`const var = require("mod").var` but `const { var } = require("mod")` is better.

decentraland/starter.js
marcelomorgado: Yeah, I've explained why on the thread https://github.com/tasitlabs/tasit/pull/207#discussion_r270443373

decentraland/App.js
marcelomorgado: Extracted to: https://github.com/tasitlabs/tasit/issues/229

pcowgill: Thanks! Resolving this

decentraland/starter.js
pcowgill: Yep, I saw that after making this comment. Resolving this

decentraland/helpers.js
pcowgill: Hmm, but here there's no `.ProviderFactory` in one version that would make this like your example, is there?

marcelomorgado: Oh, you're right.

`const { ProviderFactory } = require("tasit-action/dist/ProviderFactory");`
console.log(ProviderFactory)
// undefined

Weird... how it was working?!

pcowgill: Hmm, interesting

PR #216, opened 2019-03-27
Skip extra Apple step during build upload

pcowgill:
### Issue link

n/a

### Types of changes

### Notes

Now I won't have to answer two questions in the App Store Connect UI each time to say that the encryption our app uses qualifies for a standard exemption to their encryption rule


pcowgill:
This change worked in the latest build - I was able to skip a step on App Store Connect

PR #224, opened 2019-03-28
Back button

### LINKED ISSUE #201: Change title bar color at the top of app to match main background color
And maybe default the app in general to dark mode?

### LINKED ISSUE #202: Header back button
Since PR #198 the header is showing only the menu button but some screens should have the back button too.

Refs: https://github.com/tasitlabs/tasit/pull/198#issuecomment-476882062

marcelomorgado:
### Issue link

#202 
### Auto-close the issue?

Closes #202 
Closes #201

### Types of changes

### DIFF
--- a/decentraland/AppNavigator.js
+++ b/decentraland/AppNavigator.js
@@ -16,34 +16,64 @@ import EthereumQuestionScreen from "./screens/EthereumQuestionScreen";
 import EthereumSignUpScreen from "./screens/EthereumSignUpScreen";
 import EthereumSignInScreen from "./screens/EthereumSignInScreen";
 import MyAssetsScreen from "./screens/MyAssetsScreen";
-import Colors from "@constants/Colors.js";
 
-const defaultNavigationOptions = ({ navigation }) => {
+import Colors from "@constants/Colors.js";
+import { responsiveHeight } from "react-native-responsive-dimensions";
+
+const defaultNavigationOptions = {
+  headerStyle: {
+    backgroundColor: Colors.headerBackgroundColor,
+    height: responsiveHeight(7),
+  },
+};
+
+const headerWithMenuButton = ({ navigation }) => {
   return {
     headerLeft: (
       <Button transparent onPress={() => navigation.toggleDrawer()}>
         <Icon name="menu" />
       </Button>
     ),
-    headerStyle: {
-      backgroundColor: Colors.headerBackgroundColor,
-    },
-    headerTintColor: Colors.headerTintColor,
-    headerTitleStyle: {
-      fontWeight: "bold",
-    },
+  };
+};
+
+const headerWithBackButton = ({ navigation }) => {
+  return {
+    headerLeft: (
+      <Button transparent onPress={() => navigation.goBack()}>
+        <Icon name="md-arrow-back" />
+      </Button>
+    ),
   };
 };
 
 const AssetsForSaleNavigator = createStackNavigator(
   {
     HomeScreen,
-    ListLandForSaleScreen,
-    BuyLandScreen,
-    OnboardingHomeScreen,
-    EthereumQuestionScreen,
-    EthereumSignUpScreen,
-    EthereumSignInScreen,
+    ListLandForSaleScreen: {
+      screen: ListLandForSaleScreen,
+      navigationOptions: headerWithMenuButton,
+    },
+    BuyLandScreen: {
+      screen: BuyLandScreen,
+      navigationOptions: headerWithBackButton,
+    },
+    OnboardingHomeScreen: {
+      screen: OnboardingHomeScreen,
+      navigationOptions: headerWithBackButton,
+    },
+    EthereumQuestionScreen: {
+      screen: EthereumQuestionScreen,
+      navigationOptions: headerWithBackButton,
+    },
+    EthereumSignUpScreen: {
+      screen: EthereumSignUpScreen,
+      navigationOptions: headerWithBackButton,
+    },
+    EthereumSignInScreen: {
+      screen: EthereumSignInScreen,
+      navigationOptions: headerWithBackButton,
+    },
   },
   {
     initialRouteName: "HomeScreen",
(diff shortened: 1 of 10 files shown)


decentraland/AppNavigator.js
```diff
@@ -16,34 +16,64 @@ import EthereumQuestionScreen from "./screens/EthereumQuestionScreen";
...
 import EthereumSignInScreen from "./screens/EthereumSignInScreen";
 import MyAssetsScreen from "./screens/MyAssetsScreen";
+
 import Colors from "@constants/Colors.js";
+import { responsiveHeight } from "react-native-responsive-dimensions";
+
+const defaultNavigationOptions = {
+  headerStyle: {
+    backgroundColor: Colors.headerBackgroundColor,
+    height: responsiveHeight(7),
```
pcowgill: Note that we'll want to see how this looks on an iPhone X where there's the weird tab at the top (my personal device in an iPhone X)

decentraland/AppNavigator.js
```diff
@@ -16,34 +16,64 @@ import EthereumQuestionScreen from "./screens/EthereumQuestionScreen";
...
-      fontWeight: "bold",
-    },
+  };
+};
+
+const headerWithBackButton = ({ navigation }) => {
+  return {
+    headerLeft: (
+      <Button transparent onPress={() => navigation.goBack()}>
+        <Icon name="md-arrow-back" />
```
pcowgill: Should we have it say the word `Back` too in addition to the icon?

decentraland/AppNavigator.js
```diff
@@ -16,34 +16,64 @@ import EthereumQuestionScreen from "./screens/EthereumQuestionScreen";
...
-    EthereumQuestionScreen,
-    EthereumSignUpScreen,
-    EthereumSignInScreen,
+    ListLandForSaleScreen: {
+      screen: ListLandForSaleScreen,
+      navigationOptions: headerWithMenuButton,
+    },
+    BuyLandScreen: {
+      screen: BuyLandScreen,
+      navigationOptions: headerWithBackButton,
```
pcowgill: Oh yeah this seems like a relatively straightforward technique - nicely done!

decentraland/AppNavigator.js
```diff
@@ -16,34 +16,64 @@ import EthereumQuestionScreen from "./screens/EthereumQuestionScreen";
...
-      backgroundColor: Colors.headerBackgroundColor,
-    },
-    headerTintColor: Colors.headerTintColor,
-    headerTitleStyle: {
-      fontWeight: "bold",
-    },
+  };
+};
+
+const headerWithBackButton = ({ navigation }) => {
```
pcowgill: I wonder if we want to do an iOS/Android device check and not have the back button on Android?

pcowgill: That's complicated so would come in a future PR if we went that route

decentraland/AppNavigator.js
```diff
@@ -61,29 +91,19 @@ const MyAssetsNavigator = createStackNavigator(
...
     },
   },
-  {
-    drawerOpenRoute: "DrawerOpen",
-    drawerCloseRoute: "DrawerClose",
-    drawerToggleRoute: "DrawerToggle",
+  MyAssets: {
+    screen: MyAssetsNavigator,
     navigationOptions: {
-      drawerLockMode: "locked-closed",
```
pcowgill: Is this option the default? Or do we want to default to whatever else the default is now that the drawer menu isn't always shown?

decentraland/constants/Colors.js
```diff
@@ -13,7 +13,7 @@ export default {
   noticeText: "#fff",
   backgroundColor: "#fff",
   textColor: "gray",
-  headerBackgroundColor: "gray",
+  headerBackgroundColor: "#fff",
   headerTintColor: "#fff",
```
pcowgill: We could even DRY up these color constants by having `const WHITE = #FFF`

decentraland/constants/Colors.js
```diff
@@ -13,7 +13,7 @@ export default {
   noticeText: "#fff",
   backgroundColor: "#fff",
   textColor: "gray",
```
pcowgill: This could be `textColor: SECONDARY_APP_COLOR` with `const SECONDARY_APP_COLOR = "gray"` or something like that

decentraland/AppNavigator.js
marcelomorgado: Ok, here the headers' height was decreased a little.

decentraland/AppNavigator.js
marcelomorgado: I've read this on my mobile Gmail app and it's the same their (without text) I think that is standard enough?

decentraland/AppNavigator.js
marcelomorgado: On Android Gmail app, the back button (arrow to left) is shown
WhatsApp Image 2019-03-29 at 9 41 13 AM

decentraland/AppNavigator.js
marcelomorgado: Yes, they are default values (I've forget to delete these before creating the PR). These options aren't mentioned on `react-navigation` doc and seem that are used only for customization needs.
Refs:
https://reactnavigation.org/docs/en/drawer-navigator.html
https://github.com/react-navigation/react-navigation/issues/3266#issue-288153811

decentraland/constants/Colors.js
marcelomorgado: Done. And I've also removed unused colors.

decentraland/constants/Colors.js
marcelomorgado: Done. And I've also removed unused colors.

decentraland/AppNavigator.js
pcowgill: Resolving this

decentraland/constants/Colors.js
pcowgill: Thanks! Resolving

decentraland/constants/Colors.js
pcowgill: Thanks! Resolving

decentraland/AppNavigator.js
pcowgill: It looks like drawerLockMode is mentioned in their docs, anyway

pcowgill: Good to know they’re defaults - thanks for looking into this. Resolving!

decentraland/AppNavigator.js
pcowgill: I would argue that if we want a really mainstream audience we need to reach people who use the default email client on their device (I’d have to check how the back button looks there though).

Airbnb almost always supplements their icons with text in case people would get confused.

But either way we’d need designs for how this looks, and it’s far from the top priority, so I’ll keep it in mind and maybe create an issue for it

pcowgill: Resolving this

decentraland/AppNavigator.js
pcowgill: Oh okay, cool. Thanks for checking that.

If that’s there on Android in that app then it shouldn’t be too jarring to Android users. Resolving this!

decentraland/AppNavigator.js
pcowgill: I’ll resolve this thread and open an issue if anything looks off on iOS

PR #228, opened 2019-03-29
Show info about account creation steps to user

### LINKED ISSUE #213: Show account status text on buy screen that persists after each toast so the user knows what they're waiting for even if they miss the toast

marcelomorgado:
### Issue link

#213 
### Auto-close the issue?

Closes #213 

### Types of changes


decentraland/components/presentational/BuyLand.js
```diff
@@ -1,38 +1,60 @@
...
+    fundedWithEthers,
+    fundedWithMana,
+    approvedMarketplace,
+  } = accountInfo;
+
+  let waitingMessage = "";
+  if (waitingForAccountSetup) {
+    if (account == null) waitingMessage = "Generating account...";
+    else if (!fundedWithEthers) waitingMessage = "Funding account with ETH...";
+    else if (!fundedWithMana && !approvedMarketplace)
```
pcowgill: @marcelomorgado I think it may end up making our code easy to reason about if we move a single enumerable list of strings piece of state through states rather than dealing with the combinatorial complexity of all the booleans

marcelomorgado:
I've changed back `start.js` to ES5 because it stops working after last changes (console became silent).

decentraland/components/presentational/BuyLand.js
marcelomorgado: Hmm, create a new state var with the message and keep the booleans or delete the booleans?
I've created these flags thinking on the future screens with checks for each step.

pcowgill: The enumerable list could be generated from the booleans still, but then when showing the messages we could use the enumerable list var

pcowgill: Actually I take that back - I think the best options are either to store state in redux as the enumerable list (my favorite option) or keep it as-is.

decentraland/components/presentational/BuyLand.test.js
```diff
@@ -4,18 +4,115 @@ import BuyLand from "./BuyLand";
...
+        )
+      ).toMatchSnapshot();
+    });
+
+    it("waiting for setup - waiting for MANA funding", async () => {
+      const accountInfo = {
+        setupInProgress: true,
+        fundedWithEthers: true,
+        fundedWithMana: false,
+        approvedMarketplace: true,
```
pcowgill: I guess the fact that this can happen in either order with the MANA funding is the only potential hiccup for the enumerable list of strings idea

decentraland/components/presentational/__snapshots__/BuyLand.test.js.snap
```diff
@@ -41,3 +41,261 @@ exports[`BuyLand renders the component 1`] = `
...
+  </Component>
+  <Component
+    style={
+      Object {
+        "flexDirection": "row",
+      }
+    }
+  >
+    <Component>
+      Generating account...
```
pcowgill: If the name of this test is waiting for ETH funding, wouldn't that imply that the account has already been generated and we're on the next step of waiting for ETH funding?

pcowgill: I think maybe we need a bool for setup in progress and now account generation is complete

decentraland/components/presentational/__snapshots__/BuyLand.test.js.snap
```diff
@@ -41,3 +41,261 @@ exports[`BuyLand renders the component 1`] = `
...
```
pcowgill: Don't we want this to be something dynamic, like showing "Waiting for MANA funding" or whatever the toast says?

decentraland/components/presentational/BuyLand.js
marcelomorgado: That makes sense?
```javascript
accountCreationStatus = ['NONE','GENERATING_ACCOUNT', 
'FUNDING_WITH_ETH', 
'FUNDING_WITH_MANA_AND_APPROVING_MARKETPLACE',
'FUNDING_WITH_MANA',
'APPROVING_MARKETPLACE',
'READY_TO_USE']
```

decentraland/screens/EthereumSignUpScreen.js
```diff
@@ -39,7 +66,7 @@ export class EthereumSignUpScreen extends React.Component {
   _onSignUp = () => {
     const { setSetupInProgress } = this.props;
     setSetupInProgress(true);
-    showInfo(`Creating and funding account...`);
```
pcowgill: Is there no equivalent for this one anymore?

decentraland/components/presentational/BuyLand.test.js
pcowgill: Following the new convention, "this" here refers to the lowest line of code shown, the `approvedMarketplace` one

decentraland/screens/EthereumSignUpScreen.js
marcelomorgado: I've removed it because the toast now is being shown when some action was completed and the text on the screen is being used to show when something is processing.

decentraland/components/presentational/BuyLand.test.js
marcelomorgado: The parallelism adds more complexity, because of that I've created the message "Funding account with MANA and approving Marketplace..."

pcowgill: That makes sense. We can resolve this then.

decentraland/screens/EthereumSignUpScreen.js
pcowgill: Ah ok, so for this one it would just be "buy" clicked, and that would be silly? Got it.

decentraland/components/presentational/BuyLand.js
pcowgill: Yeah, and `FUNDING_WITH_MANA_AND_APPROVING_MARKETPLACE` can transition to `FUNDING_WITH_MANA` or `APPROVING_MARKETPLACE`? That sounds great! I think this will be a big improvement in "easiness to reason about and ability to avoid bugs"

pcowgill:
> I've changed back `start.js` to ES5 because it stops working after last changes (console became silent).

Ah ok, good to know. Yeah, we need that console output. Sounds good for now. Thanks!

decentraland/components/presentational/BuyLand.js
pcowgill: And to clarify, there would be no more account creation status booleans in redux after making this change, right? Just this one var?

marcelomorgado: > Yeah, and `FUNDING_WITH_MANA_AND_APPROVING_MARKETPLACE` can transition to `FUNDING_WITH_MANA` or `APPROVING_MARKETPLACE`? That sounds great! I think this will be a big improvement in "easiness to reason about and ability to avoid bugs"

We can pick one of both this big state is to cover the parallelism.

marcelomorgado: > And to clarify, there would be no more account creation status booleans in redux after making this change, right? Just this one var?

Yes, it can be done (only the ENUM state) for sure.

decentraland/components/presentational/__snapshots__/BuyLand.test.js.snap
marcelomorgado: The test case was buggy. Fixed.

decentraland/components/presentational/__snapshots__/BuyLand.test.js.snap
marcelomorgado: The test case was buggy. Fixed.

decentraland/screens/EthereumSignUpScreen.js
marcelomorgado: Just after clicking the button "Continue" the user will back to "BuyLand" screen and the message "Generating account..." will be shown.

pcowgill: Okay, sounds good! Resolving this.

decentraland/components/presentational/__snapshots__/BuyLand.test.js.snap
pcowgill: Thanks! Resolving this

pcowgill: Are you sure it was the test case that was buggy? Because I think this is also what I saw as a user

decentraland/components/presentational/__snapshots__/BuyLand.test.js.snap
pcowgill: Thanks! Resolving this

decentraland/components/presentational/BuyLand.js
pcowgill: > We can pick one of both this big state is to cover the parallelism.

Yep, good call.

> Yes, it can be done (only the ENUM state) for sure.

Okay, awesome!

pcowgill: Should we resolve this?

decentraland/components/presentational/__snapshots__/BuyLand.test.js.snap
marcelomorgado: Both were. The test suite wasn't looking to redux `account` prop (that is used to know if the account was generated [!= null]). And the app too because the `setAccount` was being called at the end of the setup process. The output of both (test and app) were the same.

decentraland/components/presentational/BuyLand.js
marcelomorgado: I'll do this change yet.

marcelomorgado: So, to clarify, 
1. We'll move all these props `setupInProgress,fundedWithEthers,fundedWithMana,marketplaceApproved` to `creationStatus` that will be a enum of strings.
and
2. For the paralellism, we'll assume that the `Mana funding` step occurs before `Marketplace approval` for the user feedback.

It these right?

decentraland/components/presentational/BuyLand.test.js
```diff
@@ -25,8 +26,28 @@ describe("BuyLand", () => {
...
+        fundedWithEthers: false,
+        fundedWithMana: false,
+        approvedMarketplace: false,
+      };
+      expect(
+        shallow(
+          <BuyLand
+            landForSale={parcelForSale}
+            onBuy={onBuy}
+            accountInfo={accountInfo}
```
pcowgill: Does `BuyLand` still need `accountInfo` in this format? Should we move it over to using the enum state from redux via the connect state to props in the screen that renders this presentational component?

decentraland/components/presentational/BuyLand.js
```diff
@@ -1,38 +1,60 @@
...
+      waitingMessage = "Funding account with MANA and approving Marketplace...";
```
pcowgill:
```suggestion
      waitingMessage = "Funding account with MANA and approving marketplace...";
```

decentraland/components/presentational/BuyLand.js
```diff
@@ -1,38 +1,60 @@
...
+    else if (!approvedMarketplace) waitingMessage = "Approving Marketplace...";
```
pcowgill:
```suggestion
    else if (!approvedMarketplace) waitingMessage = "Approving marketplace...";
```

decentraland/components/presentational/BuyLand.js
pcowgill: The user has no reason to know that "marketplace" is the name of the contracts, so it's more clear as a general concept if it's not capitalized

decentraland/screens/EthereumSignUpScreen.js
```diff
@@ -31,6 +31,7 @@ export class EthereumSignUpScreen extends React.Component {
 
       const account = await createAccount();
       showInfo(`Account generated`);
+      setAccount(account);
```
pcowgill: Ah ok, so this is the bug fix

decentraland/components/presentational/BuyLand.js
```diff
@@ -1,38 +1,60 @@
...
+import { StyleSheet, View, Text } from "react-native";
 import { responsiveHeight } from "react-native-responsive-dimensions";
 import PropTypes from "prop-types";
 import LandForSale from "./LandForSale";
 import Button from "./Button";
 import Colors from "@constants/Colors";
 
 export default function BuyLand(props) {
-  const { landForSale, onBuy, waitingForAccountSetup } = props;
+  const { landForSale, onBuy, accountInfo } = props;
```
pcowgill: Here is where we would change what is passed into this component https://github.com/tasitlabs/tasit/pull/228/files#r270520447

decentraland/components/presentational/BuyLand.test.js
marcelomorgado: Yes, I'll do that.

decentraland/screens/EthereumSignUpScreen.js
marcelomorgado: Yes

decentraland/components/presentational/__snapshots__/BuyLand.test.js.snap
pcowgill: Ah right, got it. Thanks for the explanation!

decentraland/screens/EthereumSignUpScreen.js
pcowgill: Cool. Resolving this

decentraland/components/presentational/BuyLand.test.js
pcowgill: Okay, sounds good!

decentraland/components/presentational/BuyLand.js
pcowgill: I agree with 1!

As for 2:

> For the parallelism, we'll assume that the `Mana funding` step occurs before `Marketplace approval` for the user feedback.

I think for user feedback we can trigger either state transition in redux and show whichever message appears first.

pcowgill: > We can pick one of both this big state is to cover the parallelism.

Oh, maybe I didn't understand this point. I don't think I agree - I'll try to diagram this.

pcowgill: Oh nevermind, I do agree that's the way to go for now. I'll show a diagram of how I think it can work once we're using event listeners in addition to shipping the two transactions in parallel

pcowgill: Frame

marcelomorgado: Got it. I've just committed the redux state change.

decentraland/components/presentational/BuyLand.test.js
marcelomorgado: Resolving this.

decentraland/components/presentational/BuyLand.test.js
marcelomorgado: Done.

pcowgill: Thanks! Resolving

decentraland/components/presentational/BuyLand.js
```diff
@@ -5,25 +5,41 @@ import PropTypes from "prop-types";
...
-    account,
-    setupInProgress: waitingForAccountSetup,
-    fundedWithEthers,
-    fundedWithMana,
-    approvedMarketplace,
-  } = accountInfo;
+  const { landForSale, onBuy, accountCreationStatus } = props;
+  const waitingForAccountSetup =
+    accountCreationStatus !== NOT_STARTED &&
+    accountCreationStatus != READY_TO_USE;
```
pcowgill: We should stick with both `!==` or both `!=`

decentraland/components/presentational/BuyLand.js
pcowgill: Oh looking at your code, I see how even without events we still are able to react to each of the two tx done in parallel separately, so it is like the diagram above, right?

decentraland/redux/reducers.js
```diff
@@ -5,43 +5,31 @@ import {
...
-    case SET_SETUP_IN_PROGRESS:
-      return { ...state, setupInProgress };
-    case SET_ACCOUNT_FUNDED_WITH_ETHERS:
-      return { ...state, fundedWithEthers };
-    case SET_ACCOUNT_FUNDED_WITH_MANA:
-      return { ...state, fundedWithMana };
-    case SET_ACCOUNT_APPROVED_MARKETPLACE:
-      return { ...state, approvedMarketplace };
+    case SET_ACCOUNT_CREATION_STATUS:
+      return { ...state, creationStatus };
```
pcowgill: Would this be where we would want to make a future PR enforcing rules about allowed state transitions?

decentraland/components/presentational/BuyLand.js
pcowgill:
```suggestion
    accountCreationStatus !== READY_TO_USE;
```

decentraland/components/presentational/BuyLand.js
pcowgill: Resolving this

decentraland/redux/reducers.js
pcowgill: @marcelomorgado Curious for your thoughts on this when you get a chance. Thanks!

decentraland/components/presentational/BuyLand.js
pcowgill: I think we can resolve this because it's now like the diagram above

decentraland/redux/reducers.js
marcelomorgado: Hey @pcowgill, it isn't an easy question, I've read a bit about it (where put business logic code on react apps). 
Using `redux` we have some options for it: `actions`, `reducers`, `thunk`/`saga`, our own `middlewares` or a specific lib, each one with pros/cons. The app scale planning seems the major concern on how the solution to use. For our current needs I think that we can use `actions` or `reducers` to that and move to `thunk`/`saga` if/when will add it to the project.

I haven't found articles about business logic using `React Context` so I need to look deep into this to have an answer.

Refs:
https://redux.js.org/faq/code-structure#how-should-i-split-my-logic-between-reducers-and-action-creators-where-should-my-business-logic-go
https://medium.com/@jeffbski/where-do-i-put-my-business-logic-in-a-react-redux-application-9253ef91ce1

decentraland/components/presentational/BuyLand.js
marcelomorgado: Yep

pcowgill: Thanks. Resolving

decentraland/redux/reducers.js
pcowgill: Thanks for looking into this! That all makes sense.

I think even if we used `thunk` or `saga` it could still make sense to enforce state transitions based on the returned data outside the scope of the async dispatch that the `thunk` or `saga` would take over.

Your suggestion of how to handle it for now sounds good. In particular I think it makes sense in the reducer based on this wording `"Reducers specify how the application's state changes in response to actions sent to the store"` ([Source](https://redux.js.org/basics/reducers))

Yeah, we can think about how `React Context` fits in maybe for `0.2.0`

Custom middleware or an open-source middleware lib might make sense if we ever want to start enforcing the same transition rules on multiple reducers

pcowgill: So long story short, I think my personal answer to my original question

> Would this be where we would want to make a future PR enforcing rules about allowed state transitions?

is yes. Does that work for you?

marcelomorgado: Yes, I agree :+1:

pcowgill: Okay, I'll resolve this and create an issue for this.

pcowgill: https://github.com/tasitlabs/tasit/issues/239

PR #237, opened 2019-04-01
Removed duplicate loading screen

### LINKED ISSUE #166: Get rid of loading screen since it's serving a duplicate role with the Expo splash screen
Modify the splash screen to have the name of the app instead

### LINKED ISSUE #217: When My Assets screen has no assets, vertically center message

pcowgill:
### Issue link

https://github.com/tasitlabs/tasit/issues/166
https://github.com/tasitlabs/tasit/issues/217

### Auto-close the issue?

Closes #166 
Closes #217 

### Types of changes


pcowgill:
@marcelomorgado This might be a `WIP` still, because right now after loading the app locally on this feature branch I'm just seeing an empty list of land for sale

marcelomorgado:
> @marcelomorgado This might be a `WIP` still, because right now after loading the app locally on this feature branch I'm just seeing an empty list of land for sale

Yeah, I've worked on that last week and not figured out why it's happening yet.

pcowgill:
@marcelomorgado I tracked down the error:

Screenshot 2019-04-02 09 41 41
Screenshot 2019-04-02 09 41 52

pcowgill:
@marcelomorgado So I think maybe after reworking the `AppNavigator` file we're not properly wrapping the app with `Root` for `native-base` to work?

pcowgill:
@marcelomorgado Wrapping the `ListLandForSaleScreen` in `Root` from `native-base` addresses the issue, for the record. But that's not the most elegant solution, since using the `Root` from `App.js` should be able to work if we set it up right.

pcowgill:
@marcelomorgado Thoughts on whether we should merge in its current form with the `Root` hack after removing `console.log`s, or try to fix that first?

marcelomorgado:
Hey @pcowgill , I've found this comment https://github.com/wix/react-native-navigation/issues/4640#issuecomment-460398078 event it isn't totally clear for me, seems that the `Root` component will not work from `App` (we can remove it). I've quickly tried to add it programmatically inside of `AppNavigator` but my attempts until now don't work. I think will be better do a comment about it and go ahead. Make sense?

marcelomorgado:
> Hey @pcowgill , I've found this comment [wix/react-native-navigation#4640 (comment)](https://github.com/wix/react-native-navigation/issues/4640#issuecomment-460398078) event it isn't totally clear for me, seems that the `Root` component will not work from `App` (we can remove it). I've quickly tried to add it programmatically inside of `AppNavigator` but my attempts until now don't work. I think will be better do a comment about it and go ahead. Make sense?

If you agree with me, this PR is ready to merge.

decentraland/App.js
```diff
@@ -11,7 +11,6 @@ import { Action } from "tasit-sdk";
 const { ConfigLoader } = Action;
 import tasitSdkConfig from "./config/default";
 import { checkBlockchain, showFatalError, recoverAccount } from "./helpers";
-import { Root } from "native-base";
```
pcowgill: @marcelomorgado Hmm, are you sure removing this won't break the toasts in the account creation / funding flow? I can test that.

pcowgill:
> Hey @pcowgill , I've found this comment [wix/react-native-navigation#4640 (comment)](https://github.com/wix/react-native-navigation/issues/4640#issuecomment-460398078) event it isn't totally clear for me, seems that the `Root` component will not work from `App` (we can remove it). I've quickly tried to add it programmatically inside of `AppNavigator` but my attempts until now don't work. I think will be better do a comment about it and go ahead. Make sense?

Hmm, but for `react-native-navigation` as opposed to `react-navigation`, the navigation is using native code and not JavaScript. We're using JavaScript though for navigation since we're using `react-navigation`. So I don't think the wix (maker of `react-native-navigation`) issues are a good reference point for this topic.

decentraland/App.js
marcelomorgado: You are right! Just undo the removal.

pcowgill:
> I've quickly tried to add it programmatically inside of `AppNavigator` but my attempts until now don't work. I think will be better do a comment about it and go ahead. Make sense?

> If you agree with me, this PR is ready to merge.

I'm on board with the idea of merging with a temporary working solution and a comment, but I'm not sure if removing Root from App.js will introduce any bugs, given that the issue was a `wix` one.

decentraland/App.js
pcowgill: Okay, thanks. I'll confirm those toasts are working in this branch anyway before merging after you added it back, since this is the branch that started messing with toasts even when Root was still in App.js

marcelomorgado:
Oh, thanks for the clarification, I got confused a little. I've changed my previous comment about the issue on the code.

decentraland/App.js
pcowgill: I'm glad I checked. Those aren't working on any other screens now. So we'll have to add the Root change on every screen that uses a `showInfo`, etc. function if we want the toasts everywhere we used to have them.

I'd be potentially okay with removing them from the account creation and funding flow since there's the on-screen text now too, but we at least need the toasts for when the land purchase is complete to work. Maybe that happens back on the `ListLandForSaleScreen` though where we already have it now?

marcelomorgado: I've tested it before put `Root` again on the `App` component. I think that is desirable be able to create a toast from every component that we want, I don't have a strong opinion about keeping or removing account creation steps toasts.

I have to check if the toast from account creation flow will be shown on list land screen even without `Root` on the `App` component. Was that your question?

pcowgill: I tested it after you put `Root` back on the `App` component, and there were still no toasts during account creation / funding (which makes sense for the same reason there were no toasts on the `ListLandForSaleScreen` in this branch until we added `Root` to that screen.)

pcowgill: > I think that is desirable be able to create a toast from every component that we want

I agree that being able to create a toast from any component we want is desirable.

> I don't have a strong opinion about keeping or removing account creation steps toasts.

Good to know you don't have a strong opinion about the account creation and funding flow toasts.

> I have to check if the toast from account creation flow will be shown on list land screen even without Root on the App component. Was that your question?

I'm not wondering about that, no.

marcelomorgado: Hmmmmm got it! sorry for didn't test it with more care!
As you said, we can add `<Root>` on every component that is using Toast/NativeBase or I can spend more time trying to fix it properly. What do you prefer?

pcowgill: No worries! I'm fine with adding `Root` on every component for now and creating an issue.

marcelomorgado: Seems that adding it on all components isn't a option:
```sh

[20:35:22] Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.%s, 
    in ToastContainer (at connectStyle.js:384)
    in Styled(ToastContainer) (at Root.js:15)
    in RCTView (at View.js:44)
    in Root (at connectStyle.js:384)
    in Styled(Root) (at BuyLandScreen.js:111)
    in BuyLandScreen (created by Context.Consumer)
    in Connect(BuyLandScreen) (created by SceneView)
    in SceneView (at StackViewLayout.js:782)
    in RCTView (at View.js:44)
    in RCTView (at View.js:44)
    in RCTView (at View.js:44)
    in AnimatedComponent (at StackViewCard.js:69)
    in RCTView (at View.js:44)
    in AnimatedComponent (at screens.native.js:59)
    in Screen (at StackViewCard.js:57)
    in Card (at createPointerEventsContainer.js:27)
    in Container (at StackViewLayout.js:858)
```

marcelomorgado: I'll spend more time trying to solve that on the properly way, do you agree?

pcowgill: Sure, feel free to do a deep dive on this. It seems like it must have something to do with nested navigators, but it so I can’t think of why it didn’t surface during your PR that introduced those

marcelomorgado: An update: the [`NativeBase` + `ReactNavigation` example](https://docs.nativebase.io/docs/examples/navigation/StackNavigationExample.html) works in some way like our `HomeScreen` (a screen with `this.props.navigation.navigate()` call), 
I didn't find more info about that integration. As a workaround solution I think that we can get back with `HomeScreen` but navigating to the `ListLand` without a timeout, or Create our own `SideMenu` component like the example.

pcowgill: Weird, they don't use `Root` in that example at all, do they?

It sure seems like since our `ListLand` screen is another screen that does some navigate(), we ought to be able to get that screen to work as the `HomeScreen`. Keeping a `HomeScreen` screen that does nothing is fine as a hack, but it's not ideal as a demo. Same goes for a custom `SideMenu` - if that seems easiest, feel free to try it, but it doesn't seem like it should be necessary.

Maybe we can think of some content we'd like to show the user as a means of "coaching" them before showing the `ListLand` screen anyway?

Thanks for looking into this - this is trickier than it would have seemed at first glance!

pcowgill: I just checked, and their `react-navigation` version in that example is 1, and we're on version 3. So maybe we shouldn't use that as a reference.

marcelomorgado: I've searched again using the error message and seems that is a `NativeBase` bug 
https://github.com/GeekyAnts/NativeBase/issues/1790#issuecomment-477035352

pcowgill: Hmm okay, interesting. What do you think we should do in the meantime?

pcowgill: > Maybe we can think of some content we'd like to show the user as a means of "coaching" them before showing the `ListLand` screen anyway?

@marcelomorgado Maybe this is the way to go?

marcelomorgado: Yes, we have few options now and that seems the better one. Any idea of the content that this screen will have?

pcowgill: > I'd be potentially okay with removing them from the account creation and funding flow since there's the on-screen text now too, but we at least need the toasts for when the land purchase is complete to work. Maybe that happens back on the ListLandForSaleScreen though where we already have it now?

@marcelomorgado Oh actually, if we merge it as-is, the toasts for finishing buying a piece of land do show up back on the list land screen, so we would have all the toasts we absolutely need if we were to merge this with `Root` just on the `ListLand` screen and in `App.js`, which leads to no errors. Thoughts?

marcelomorgado: Unfortunately, the toast will stop to work on all screen as before, but it's a good option too since we can merge this with no terrible side effects and work on the "coaching" screen and have the `toast` back to all screens soon.

pcowgill: I see you gave a thumbs up in response to the last message, so I will go forward with the merging as-is plan for now.

pcowgill:
@marcelomorgado Can you review and approve the latest changes before we merge this? Thanks!

pcowgill:
(Not much has changed - just to follow our usual process)

decentraland/App.js
pcowgill: > Unfortunately, the toast will stop to work on all screen as before, but it's a good option too since we can merge this with no terrible side effects and work on the "coaching" screen and have the `toast` back to all screens soon.

Okay, sounds good to me.

PR #248, opened 2019-04-03
Making publish a separate step

pcowgill:
### Issue link

n/a

### Types of changes


pcowgill:
@marcelomorgado FYI the updated version code from app.json doesn't make it into the Android build _unless_ you publish first. That was surprising. So maybe we want to revert this, at least on Android. The same may be true for iOS.

marcelomorgado:
> @marcelomorgado FYI the updated version code from app.json doesn't make it into the Android build _unless_ you publish first. That was surprising. So maybe we want to revert this, at least on Android. The same may be true for iOS.

OK, I've found these links that maybe help to understand why it's happening:
https://github.com/expo/expo/issues/63
https://forums.expo.io/t/how-to-build-without-publishing/14462

pcowgill:
So if I understand the forum post correctly, the suggestion is to publish first to a non-prod release channel to get the build to increment the version code in the .apk, and then follow these instructions to move that published js bundle to a different release channel.

https://docs.expo.io/versions/v30.0.0/distribution/advanced-release-channels/#promoting-a-release-to-a-new-channel

pcowgill:
If that's the recommendation, I should probably remove the `--no-publish` flag from the scripts

PR #251, opened 2019-04-03
Optimistic update

### LINKED ISSUE #214: Land parcel should be removed from assets for sale list in redux immediately after triggering the buy and added back if the buy fails to go through
(In other words, optimistic updates)

marcelomorgado:
### Issue link

#214 

### Auto-close the issue?

Closes #214 

### Types of changes

### DIFF
--- a/decentraland/redux/reducers.js
+++ b/decentraland/redux/reducers.js
@@ -8,6 +8,7 @@ import {
   SET_ACCOUNT_CREATION_STATUS,
   SET_LOADING_ASSETS_FOR_SALE_IN_PROGRESS,
   ADD_TO_MY_ASSETS_LIST,
+  REMOVE_MY_ASSET_FROM_LIST,
 } from "./actions";
 
 import AccountCreationStatus from "@constants/AccountCreationStatus";
@@ -66,6 +67,8 @@ function myAssets(state = { list: [] }, action) {
   switch (type) {
     case ADD_TO_MY_ASSETS_LIST:
       return { ...state, list: [myAsset, ...state.list] };
+    case REMOVE_MY_ASSET_FROM_LIST:
+      return { ...state, list: state.list.filter(val => val !== myAsset) };
     default:
       return state;
   }
--- a/decentraland/redux/reducers.js
+++ b/decentraland/redux/reducers.js
@@ -2,7 +2,6 @@ import { combineReducers } from "redux";
 import {
   SET_ACCOUNT,
   SELECT_LAND_TO_BUY,
-  SET_LAND_FOR_SALE_LIST,
   REMOVE_LAND_FOR_SALE,
   ADD_LAND_FOR_SALE_TO_LIST,
   SET_ACCOUNT_CREATION_STATUS,
@@ -47,14 +46,16 @@ function selectedLandToBuy(state = null, action) {
 }
 
 function assetsForSale(state = { list: [], loadingInProgress: true }, action) {
-  const { type, list, landForSale, loadingInProgress } = action;
+  const { type, landForSale, loadingInProgress } = action;
   switch (type) {
-    case SET_LAND_FOR_SALE_LIST:
-      return { ...state, list };
     case ADD_LAND_FOR_SALE_TO_LIST:
-      return { ...state, list: [...state.list, landForSale] };
-    case REMOVE_LAND_FOR_SALE:
-      return { ...state, list: state.list.filter(val => val !== landForSale) };
+      return { ...state, list: [landForSale, ...state.list] };
+    case REMOVE_LAND_FOR_SALE: {
+      let { list: assetsForSale } = state;
+      const toRemove = landForSale;
+      const list = assetsForSale.filter(asset => asset !== toRemove);
+      return { ...state, list };
+    }
     case SET_LOADING_ASSETS_FOR_SALE_IN_PROGRESS:
       return { ...state, loadingInProgress };
     default:
@@ -67,8 +68,12 @@ function myAssets(state = { list: [] }, action) {
   switch (type) {
     case ADD_TO_MY_ASSETS_LIST:
       return { ...state, list: [myAsset, ...state.list] };
-    case REMOVE_MY_ASSET_FROM_LIST:
-      return { ...state, list: state.list.filter(val => val !== myAsset) };
+    case REMOVE_MY_ASSET_FROM_LIST: {
+      const { list: myAssets } = state;
+      const toRemove = myAsset;
+      const list = myAssets.filter(asset => asset !== toRemove);
+      return { ...state, list };
+    }
     default:
       return state;
   }
--- a/decentraland/screens/BuyLandScreen.js
+++ b/decentraland/screens/BuyLandScreen.js
@@ -1,6 +1,11 @@
 import React from "react";
 import { connect } from "react-redux";
-import { removeLandForSale, addToMyAssetsList } from "../redux/actions";
+import {
+  removeLandForSale,
+  addLandForSaleToList,
+  removeMyAssetFromList,
+  addToMyAssetsList,
+} from "../redux/actions";
 import BuyLand from "@presentational/BuyLand";
 import PropTypes from "prop-types";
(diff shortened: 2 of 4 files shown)


pcowgill:
@marcelomorgado I'm going to start reviewing this now - one top of mind question before I start is whether the error path is handled. Noting that so I don't forget while reviewing it.

decentraland/redux/reducers.js
```diff
@@ -66,6 +67,8 @@ function myAssets(state = { list: [] }, action) {
   switch (type) {
     case ADD_TO_MY_ASSETS_LIST:
       return { ...state, list: [myAsset, ...state.list] };
+    case REMOVE_MY_ASSET_FROM_LIST:
+      return { ...state, list: state.list.filter(val => val !== myAsset) };
```
pcowgill: I wonder if this would be more readable split out into a couple lines of code. Certainly a matter of personal preference and not important in a vacuum, but having the style match throughout the codebase has some value for understandability. Thoughts?

decentraland/screens/BuyLandScreen.js
```diff
@@ -53,16 +60,29 @@ export class BuyLandScreen extends React.Component {
 
     const onSuccess = () => {
       showInfo(`${typeDescription} bought successfully.`);
-      removeLandForSale(landForSale);
-      addToMyAssetsList(asset);
+    };
+
+    const onError = () => {
```
pcowgill: In a future version I wonder whether we want the error payload to contain all the data we'd need to use to react to the error event (here, the asset in question)

decentraland/screens/BuyLandScreen.js
```diff
@@ -53,16 +60,29 @@ export class BuyLandScreen extends React.Component {
...
+      removeMyAssetFromList(asset);
+      addLandForSaleToList(landForSale);
```
pcowgill: It's weird that the argument for these two functions isn't the exact same object or asset id. What's the difference between the contents of these two variables?

decentraland/screens/BuyLandScreen.js
```diff
@@ -53,16 +60,29 @@ export class BuyLandScreen extends React.Component {
 
     const onSuccess = () => {
       showInfo(`${typeDescription} bought successfully.`);
-      removeLandForSale(landForSale);
-      addToMyAssetsList(asset);
+    };
+
+    const onError = () => {
+      removeMyAssetFromList(asset);
```
pcowgill: We probably want to show an alert or toast in this case

decentraland/screens/BuyLandScreen.js
```diff
@@ -53,16 +60,29 @@ export class BuyLandScreen extends React.Component {
...
+      addLandForSaleToList(landForSale);
     };
 
     showInfo(`Buying the ${typeDescription.toLowerCase()}...`);
-    _executeOrder(landForSale, account, onSuccess);
+
+    // Optimistic UI update
+    removeLandForSale(landForSale);
+    addToMyAssetsList(asset);
```
pcowgill: Maybe in a future PR, but the reducer for this action should probably set a pending state that is then set to confirmed onSuccess

decentraland/redux/reducers.js
marcelomorgado: Agree. I'll do that.

decentraland/screens/BuyLandScreen.js
marcelomorgado: The `landForSale` is the sell order and wraps the asset. For example:
```javascript
const assetForSale = {
  priceInWei,
  expiresAt,
  asset: {
    id, type, x, y
  }
}
```
The `MyAssetsScreen` has assets owned by the user instead of sell orders because it's no longer an open sell order.

decentraland/screens/BuyLandScreen.js
marcelomorgado: Yes.

decentraland/screens/BuyLandScreen.js
marcelomorgado: Extracted to: https://github.com/tasitlabs/tasit/issues/253

decentraland/screens/BuyLandScreen.js
marcelomorgado: Could you clarify your point better?

decentraland/screens/BuyLandScreen.js
pcowgill: Ah ok right, that makes sense - it seems like there may be some clever way to DRY that up a bit, but I don't have a concrete suggestion so let's keep it as-is for now. Feel free to resolve this after reading.

decentraland/screens/BuyLandScreen.js
pcowgill: Great, thanks

decentraland/screens/BuyLandScreen.js
pcowgill: Okay, great - thanks. Resolving this.

decentraland/screens/BuyLandScreen.js
pcowgill: Oh, like if this were

`onError = (asset OR landForSale) => {`

and then below we used that data that was passed into the error handler to call actions rather than variables from a more global scope.

decentraland/redux/reducers.js
marcelomorgado: Done.

marcelomorgado:
After my last change here, I've changed the `addLandForSaleToList` to add the `assetForSale` on the beginning of the list since on error case it was backing to the end of the list.
As a side effect, on the app startup, the list refreshing is reloading to the user on each new asset addition. I think it's a good change because on the app loading adding on end seems that the list is loading faster (because only the on top 3 assets are shown to the user). What do you think?

decentraland/screens/BuyLandScreen.js
marcelomorgado: Done.

decentraland/screens/BuyLandScreen.js
marcelomorgado: Done.

pcowgill: Cool, thanks! Resolving this.

decentraland/screens/BuyLandScreen.js
pcowgill: Resolving this. Thanks.

pcowgill:
> After my last change here, I've changed the `addLandForSaleToList` to add the `assetForSale` on the beginning of the list since on error case it was backing to the end of the list.
> As a side effect, on the app startup, the list refreshing is reloading to the user on each new asset addition. I think it's a good change because on the app loading adding on end seems that the list is loading faster (because only the on top 3 assets are shown to the user). What do you think?

Makes sense and sounds good to me!

decentraland/redux/reducers.js
pcowgill: Thanks! Resolving

decentraland/redux/reducers.js
```diff
@@ -67,8 +68,12 @@ function myAssets(state = { list: [] }, action) {
   switch (type) {
     case ADD_TO_MY_ASSETS_LIST:
       return { ...state, list: [myAsset, ...state.list] };
-    case REMOVE_MY_ASSET_FROM_LIST:
-      return { ...state, list: state.list.filter(val => val !== myAsset) };
+    case REMOVE_MY_ASSET_FROM_LIST: {
+      const { list: myAssets } = state;
+      const toRemove = myAsset;
+      const list = myAssets.filter(asset => asset !== toRemove);
```
pcowgill: It's probably overkill, but we could dry this logic up with the reducer for land for sale if we wanted to. Doesn't need to happen in this PR or ever, though. Just a thought.

pcowgill:
@marcelomorgado Merging this now! Great work!

decentraland/redux/reducers.js
marcelomorgado: Done on https://github.com/tasitlabs/tasit/pull/256. Resolving this.

PR #258, opened 2019-04-04
My land

### LINKED ISSUE #210: Purchased assets should show up after app reload - load “my assets” list from blockchain
For now, an asset is added to `my assets` list by `redux` after the user bought it.
Since we'll persist the ephemeral account, we should load the list every time the user starts the app.

Gated by  #190

Make sense that this data from `Decentraland API` or it should be getting from the `EstateRegistry` and `LandRegistry` contracts?

marcelomorgado:
### Issue link

#210 

### Types of changes

### Notes

- This PR is gated by PR #257 and should be merged after it.

- This PR has step 1 from #210: Store "My assets" locally.
A new PR will deal with getting assets from `DecentralandUtils` and closes #210.

### DIFF
--- a/decentraland/App.js
+++ b/decentraland/App.js
@@ -48,12 +48,12 @@ Is the 'config/default.js' file correct?`;
 
   async _loadAccount() {
     const account = await retrieveEphemeralAccount();
-    store.dispatch(setAccount(account));
+    if (account) store.dispatch(setAccount(account));
   }
 
   async _loadMyAssets() {
     const myAssets = await retrieveMyAssets();
-    store.dispatch(setMyAssetsList(myAssets));
+    if (myAssets) store.dispatch(setMyAssetsList(myAssets));
   }
 
   render() {
--- a/decentraland/App.js
+++ b/decentraland/App.js
@@ -6,11 +6,12 @@ import AppNavigator from "./AppNavigator";
 import { Provider } from "react-redux";
 import { createStore } from "redux";
 import decentralandApp from "./redux/reducers";
-import { setAccount } from "./redux/actions";
+import { setAccount, setMyAssetsList } from "./redux/actions";
 import { Action } from "tasit-sdk";
 const { ConfigLoader } = Action;
 import tasitSdkConfig from "./config/default";
-import { checkBlockchain, showFatalError, recoverAccount } from "./helpers";
+import { checkBlockchain, showFatalError } from "@helpers";
+import { retrieveEphemeralAccount, retrieveMyAssets } from "@helpers/storage";
 import { Ionicons } from "@expo/vector-icons";
 import { Root } from "native-base";
 
@@ -46,8 +47,13 @@ Is the 'config/default.js' file correct?`;
   }
 
   async _loadAccount() {
-    const account = await recoverAccount();
-    if (account != null) store.dispatch(setAccount(account));
+    const account = await retrieveEphemeralAccount();
+    store.dispatch(setAccount(account));
+  }
+
+  async _loadMyAssets() {
+    const myAssets = await retrieveMyAssets();
+    store.dispatch(setMyAssetsList(myAssets));
   }
 
   render() {
@@ -75,8 +81,9 @@ Is the 'config/default.js' file correct?`;
     const setupSDK = this._setupTasitSDK();
     const loadFonts = this._loadFonts();
     const loadAccount = this._loadAccount();
+    const loadMyAssets = this._loadMyAssets();
 
-    return Promise.all([setupSDK, loadFonts, loadAccount]);
+    return Promise.all([setupSDK, loadFonts, loadAccount, loadMyAssets]);
   };
 
   _handleLoadingError = error => {
--- a/decentraland/helpers/storage.js
+++ b/decentraland/helpers/storage.js
@@ -1,0 +1,59 @@
+import { SecureStore } from "expo";
+import { createFromPrivateKey } from "tasit-account/dist/testHelpers/helpers";
+
+// Storage keys
+const EPHEMERAL_ACCOUNT_PRIV_KEY = "EPHEMERAL_ACCOUNT_PRIV_KEY";
+const MY_ASSETS_LIST = "MY_ASSETS_LIST";
+
+const storeEphemeralAccount = async account => {
+  const { privateKey } = account;
+  await _storeData(EPHEMERAL_ACCOUNT_PRIV_KEY, privateKey);
+};
+
+const retrieveEphemeralAccount = async () => {
+  let account = null;
+
+  const privateKey = await _retrieveData(EPHEMERAL_ACCOUNT_PRIV_KEY);
+  if (privateKey != null) account = createFromPrivateKey(privateKey);
+
+  return account;
+};
+
+const storeMyAssets = async myAssets => {
+  const strMyAssests = JSON.stringify(myAssets);
+  await _storeData(MY_ASSETS_LIST, strMyAssests);
+};
+
+const retrieveMyAssets = async () => {
+  const strMyAssests = await _retrieveData(MY_ASSETS_LIST);
+  const myAssets = JSON.parse(strMyAssests);
+  return myAssets;
+};
+
+// Note: the value should be a string
+const _storeData = async (key, value) => {
+  try {
+    // More about options:
+    // https://docs.expo.io/versions/latest/sdk/securestore/#securestoresetitemasynckey-value-options
+    const options = { keychainAccessible: SecureStore.WHEN_UNLOCKED };
+    await SecureStore.setItemAsync(key, value, options);
+  } catch (error) {
+    throw Error("Unable to securely store data.");
+  }
+};
+
+const _retrieveData = async key => {
+  try {
+    const value = await SecureStore.getItemAsync(key);
+    return value;
+  } catch (error) {
+    throw Error("Unable to retrieve data from storage.");
+  }
+};
+
+export default {
+  storeEphemeralAccount,
+  retrieveEphemeralAccount,
+  storeMyAssets,
+  retrieveMyAssets,
+};
--- a/decentraland/helpers/storage.js
+++ b/decentraland/helpers/storage.js
@@ -22,13 +22,11 @@ export const retrieveEphemeralAccount = async () => {
 export const storeMyAssets = async myAssets => {
   const strMyAssets = JSON.stringify(myAssets);
   await _storeData(MY_ASSETS_LIST, strMyAssets);
-  console.log(`Stored asstes -> ${strMyAssets}`);
 };
 
 export const retrieveMyAssets = async () => {
   const strMyAssets = await _retrieveData(MY_ASSETS_LIST);
   const myAssets = JSON.parse(strMyAssets);
-  console.log(`Retrieved assets -> ${strMyAssets}`);
   return myAssets;
 };
 
--- a/decentraland/helpers/storage.js
+++ b/decentraland/helpers/storage.js
@@ -1,4 +1,5 @@
 import { SecureStore } from "expo";
+import { AsyncStorage } from "react-native";
 import { createFromPrivateKey } from "tasit-account/dist/testHelpers/helpers";
 
 // Storage keys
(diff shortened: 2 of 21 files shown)


decentraland/App.js
```diff
@@ -46,8 +47,13 @@ Is the 'config/default.js' file correct?`;
   }
 
   async _loadAccount() {
-    const account = await recoverAccount();
-    if (account != null) store.dispatch(setAccount(account));
+    const account = await retrieveEphemeralAccount();
+    store.dispatch(setAccount(account));
```
pcowgill: Noting that I need to check how we handle `null` after this change

decentraland/helpers/storage.js
```diff
@@ -0,0 +1,59 @@
...
+export const retrieveEphemeralAccount = async () => {
+  let account = null;
+
+  const privateKey = await _retrieveData(EPHEMERAL_ACCOUNT_PRIV_KEY);
+  if (privateKey != null) account = createFromPrivateKey(privateKey);
+
+  return account;
+};
+
+export const storeMyAssets = async myAssets => {
```
pcowgill: I wonder if it's overkill to encrypt the assets? Should we use [`AsyncStorage`](https://docs.expo.io/versions/latest/react-native/asyncstorage/) as opposed to [`SecureStore`](https://docs.expo.io/versions/latest/sdk/securestore/) for my assets, most likely for performance reasons?

decentraland/helpers/storage.js
```diff
@@ -0,0 +1,59 @@
...
+  const strMyAssets = JSON.stringify(myAssets);
+  await _storeData(MY_ASSETS_LIST, strMyAssets);
```
pcowgill: The next time we have to persist some 3rd thing, it may be worth considering automatically persisting and rehydrating our redux store. But I think that would be premature-abstraction (which I can sometimes be guilty of) to include in this PR.

decentraland/helpers/storage.js
```diff
@@ -0,0 +1,59 @@
...
+  let account = null;
+
+  const privateKey = await _retrieveData(EPHEMERAL_ACCOUNT_PRIV_KEY);
+  if (privateKey != null) account = createFromPrivateKey(privateKey);
+
+  return account;
+};
+
+export const storeMyAssets = async myAssets => {
+  const strMyAssets = JSON.stringify(myAssets);
```
pcowgill: If we wanted to we could have gotten pretty fine-grained with splitting out the object into keys and values and persisting them separately. Thoughts on pros/cons?

decentraland/helpers/storage.js
```diff
@@ -0,0 +1,59 @@
...
+  await _storeData(MY_ASSETS_LIST, strMyAssets);
+};
+
+export const retrieveMyAssets = async () => {
+  const strMyAssets = await _retrieveData(MY_ASSETS_LIST);
+  const myAssets = JSON.parse(strMyAssets);
```
pcowgill: Always need a try/catch with things that throw like `JSON.parse`

decentraland/redux/reducers.js
```diff
@@ -77,6 +78,10 @@ function myAssets(state = { list: [] }, action) {
       const list = removeFromList(myAssets, myAsset);
       return { ...state, list };
     }
+    case SET_MY_ASSETS_LIST: {
+      const list = myAssets === null ? [] : myAssets;
```
pcowgill: I wonder whether we want to ensure that the `myAssets` state variable as used via the UI isn't `null` in this case too - not just when hitting storage via this reducer

decentraland/screens/BuyLandScreen.js
```diff
@@ -45,11 +45,13 @@ export class BuyLandScreen extends React.Component {
     const {
       navigation,
       accountInfo,
+      myAssets,
       removeLandForSale,
       prependLandForSaleToList,
       removeMyAssetFromList,
       addToMyAssetsList,
     } = props;
+    let { removeMyAssetFromList, addToMyAssetsList } = props;
```
pcowgill: Did this need to be a `let` as opposed to a `const`?

marcelomorgado: I've thought that will be necessary to override these functions but it didn't. Just fixed.

decentraland/redux/reducers.js
marcelomorgado: Not sure if I've got it.
On the way that it is now `myAssets.list` never will be null. It starts as `[]`, both `add` and `remove` actions will keep it as an array and the `set` is checking it.
Oh, worth check if `undefined` or non-array values are being passed?

decentraland/helpers/storage.js
marcelomorgado: Agree, I know that we discussed recently (maybe yesterday hahaha) about middleware, but I think on the next storage demand it'll become necessary. Any definition which path we should go?

Refs: https://github.com/tasitlabs/tasit/pull/228#discussion_r270548313

decentraland/App.js
marcelomorgado: Any thoughts? I've removed the `if` because the default state for the account is `null`.

decentraland/helpers/storage.js
marcelomorgado: Oh, you are right. Done.

decentraland/helpers/storage.js
marcelomorgado: Done.

pcowgill: Cool, thanks. Resolving

decentraland/helpers/storage.js
pcowgill: Great, thanks. Resolving.

decentraland/App.js
pcowgill: That makes sense to me. We can go with the way you have it now. Resolving this.

decentraland/helpers/storage.js
pcowgill: Haha yep. I'm not sure which path we should go yet.

decentraland/helpers/storage.js
```diff
@@ -7,47 +8,73 @@ const MY_ASSETS_LIST = "MY_ASSETS_LIST";
...
-  await _storeData(MY_ASSETS_LIST, strMyAssets);
+  const strMyAssets = _fromJson(myAssets);
+  await _storeData(MY_ASSETS_LIST, strMyAssets, false);
 };
 
 export const retrieveMyAssets = async () => {
-  const strMyAssets = await _retrieveData(MY_ASSETS_LIST);
-  const myAssets = JSON.parse(strMyAssets);
+  const strMyAssets = await _retrieveData(MY_ASSETS_LIST, false);
+  const myAssets = _toJson(strMyAssets);
```
pcowgill: Hmm, when we're retrieving them don't we want them to become an object rather than JSON?

pcowgill: I think you have them flipped

decentraland/helpers/storage.js
```diff
@@ -7,47 +8,73 @@ const MY_ASSETS_LIST = "MY_ASSETS_LIST";
...
   return myAssets;
 };
 
+const _fromJson = json => {
+  try {
+    return JSON.stringify(json);
+  } catch {
+    throw Error(`Unable to parse JSON object to a String.`);
```
pcowgill: This would be converting an object to a JSON string

decentraland/redux/reducers.js
pcowgill: I guess my question is why `myAssets` would ever be null when passed into this reducer

decentraland/screens/BuyLandScreen.js
pcowgill: Okay great, thanks. Resolving this.

decentraland/helpers/storage.js
pcowgill: This won't gate merging

decentraland/helpers/storage.js
pcowgill: We can resolve this if you like.

decentraland/helpers/storage.js
marcelomorgado: Not strong opinion, I think that we should keep it in mind and do that when it'll help us with some feature that demands to deal with the assets in that manner.

decentraland/helpers/storage.js
marcelomorgado: Resolving this.

decentraland/redux/reducers.js
marcelomorgado: Sorry, you meant "ever" or "never"?

decentraland/helpers/storage.js
marcelomorgado: Hmmm, I'm meaning JSON = object and String = JSON object parsed to string.

marcelomorgado: Correct me if I'm wrong but for me Json (JavaScript Object Notation) is the way that JS represents their objects.
So, I've meant
```javascript
json = toJson = { a: 1, b: 2, c: 3 }
string = fromJson = '{"a":1,"b":2,"c":3}'
```

pcowgill: I think JS represents their objects in one way, and JSON is a _notation_ for representing those objects as strings in a way that as closely resembles the object representation as possible. So JavaScript objects !== JSON

decentraland/redux/reducers.js
pcowgill: I meant "ever".

In other words, I think that we should make sure that the data we pass to this reducer is either an empty array or a non-empty array, but never null.

decentraland/helpers/storage.js
pcowgill: Yep, sounds good. Resolving this!

decentraland/helpers/storage.js
marcelomorgado: :thinking: 
Hmmm... I agree with your argument but it stills confusing for me since I always read `parse` functions as something like that: `Type.parse(data) => convert data to type`, so, using that logic 
`obj = JSON.parse(string) => convert string to JSON and assign to obj` and `obj === stringifyedObject === JSON`.
Anyway, could be better rename functions to `toString/fromString` or `toObject/fromObject` since that same misconception could occur with the code readers?

pcowgill: > `obj = JSON.parse(string) => convert string to JSON and assign to obj` and `obj === stringifyedObject === JSON`.

This ^ is the part I don't think is correct. Here's my understanding (and I'm 99% sure of this):

`obj = JSON.parse(string)` => convert string to a JavaScript object, and the utility library that happens to handle this is named `JSON` (because of its ability to turn objects into JSON). Assign to `obj` and `obj !== stringifiedObject === JSON`.

pcowgill: (Note that REST APIs use JSON, and that would only be possible with strings, not objects.)

pcowgill: > Anyway, could be better rename functions to `toString/fromString` or `toObject/fromObject` since that same misconception could occur with the code readers?

This works for me!

marcelomorgado: I think that I've found the missing 1% here :smile: :
`JSON is a text syntax that facilitates structured data interchange between all programming languages.` - 1st line of ECMA-404.pdf

pcowgill: Hahaha

decentraland/helpers/storage.js
marcelomorgado: Done.

decentraland/helpers/storage.js
marcelomorgado: Done, thanks for that discussion!

decentraland/redux/reducers.js
marcelomorgado: Oh, yes, I agree. So do you think will be better to remove that validation from here?

decentraland/helpers/storage.js
pcowgill: Thanks! Resolving

decentraland/helpers/storage.js
pcowgill: Thanks to you too! Great, resolving this.

decentraland/redux/reducers.js
pcowgill: Well, it's not even validation that we're doing here as it's written. There may even be a missing `.list` based on the way you described it above?

How I read it now is, sometimes we'll pass in a null value for myAssets. If that happens, pretend we passed in [] instead. Otherwise, use myAssets (not myAssets.list)

marcelomorgado: I've putted back the checks on `App.js`, we can change `retrieveMyAssets` functions to return `[]` instead of `null` too, anyway I often put that kind of validation closest to persistance to protected data from a buggy input code.

marcelomorgado: > Well, it's not validation that we're doing here as it's written. There may be a missing `.list` based on the way you described it above?
> 
> How I read it now is, sometimes we'll pass in a null value for myAssets. If that happens, pretend we passed in [] instead. Otherwise, use myAssets (not myAssets.list)

Yes, maybe we should rename vars. The action passes an array called `myAssets`, the same name as the `reducer`, but the reducer has that structure `{ list: [] }`.
So, if action sends a null, the code will stores an `[]` instead.

pcowgill: Ah okay yeah, I think what I was wanting what changing `retrieveMyAssets` to return `[]` instead of `null`.

Now I think I get what you were saying - I wasn't always making sense in this thread. What we're doing here is enforcing valid data in persistence in a sense, and I like the idea of enforcing that at the reducer level. So I changed my mind, this line should stay as-is.

pcowgill: Or wait, retrieving the assets happens from the data store or the blockchain?

Retrieving from the datastore will automatically never be null if we enforce this on the way in.

The function we may want to change is the one that gets data from the blockchain.

marcelomorgado: retrieveMyAssets changed

pcowgill: > Retrieving from the datastore will automatically never be null if we enforce this on the way in.

This isn't totally true actually - it would be null if it has never been set.

marcelomorgado: > Or wait, the retrieving the assets done by `retrieveMyAssets` is getting it from the data store or the blockchain?
> 
> Retrieving from the datastore will automatically never be null if we enforce this on the way in.
> 
> The function we may want to change is the one that gets data from the blockchain.

retrieveMyAssets is getting data from store, if has no store yet it was returning `null`, now `[]`.
I'll make that caution on the next PR (when I'll get data from `DecentralandUtils`)

pcowgill: > The function we may want to change is the one that gets data from the blockchain.

This one should return `[]` when there are no assets. But that's coming up in the next PR

pcowgill: All your responses make sense and sound good. We should rename `retrieveMyAssets` to indicate it's retrieving them from storage too.

pcowgill: I'm going to merge this PR as-is

PR #189, opened 2019-03-22
UI improvements

marcelomorgado:
### Issue link
#134 

### Auto-close the issue?

Closes #134 

### Types of changes

Technical debt (a code change that doesn't fix a bug or add a feature but makes something clearer for devs)

PR #259, opened 2019-04-04
Load account assets from blockchain

### LINKED ISSUE #210: Purchased assets should show up after app reload - load “my assets” list from blockchain

marcelomorgado:
### Issue link

#210 
https://github.com/tasitlabs/tasit/pull/189#discussion_r268775402

### Auto-close the issue?

Closes #210 

### Types of changes

### DIFF
--- a/decentraland/helpers/decentraland.js
+++ b/decentraland/helpers/decentraland.js
@@ -143,7 +143,7 @@ const _getType = (nftAddress, landContract, estateContract) => {
 
   if (!isParcel && !isEstate)
     throw new Error(
-      `The sell order should have a parcel of land or an estate as NFT.`
+      `The sell order should have a parcel of land or an estate as the NFT.`
     );
 
   if (isEstate) return ESTATE;
--- a/decentraland/helpers/decentraland.js
+++ b/decentraland/helpers/decentraland.js
@@ -4,17 +4,17 @@ import DecentralandUtils from "tasit-sdk/dist/helpers/DecentralandUtils";
 import AssetTypes from "@constants/AssetTypes";
 const { ESTATE, PARCEL } = AssetTypes;
 
-export const loadAssetsForSale = async appendLandForSaleToList => {
+export const loadAssetsForSale = async appendAssetForSaleToList => {
   const decentralandUtils = new DecentralandUtils();
   const { getAllAssetsForSale } = decentralandUtils;
 
-  const openSellOrdersEvents = await getAllAssetsForSale();
+  const openSellOrders = await getAllAssetsForSale();
 
   let contracts = getContracts();
   const { estateContract } = contracts;
 
   const estatesForSale = [];
-  for (let order of openSellOrdersEvents) {
+  for (let order of openSellOrders) {
     const { nftAddress } = order;
     const isEstate = addressesAreEqual(nftAddress, estateContract.getAddress());
     if (isEstate) estatesForSale.push(order);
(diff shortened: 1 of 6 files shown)


decentraland/helpers/decentraland.js
```diff
@@ -0,0 +1,156 @@
+// Note: Functions that will deal with fetch and prepare data from Decentraland contracts
+import { addressesAreEqual, getContracts } from ".";
```
pcowgill: Worth doing `@helpers` anyway?

decentraland/screens/MyAssetsScreen.js
```diff
@@ -1,29 +1,63 @@
...
+  account: PropTypes.object,
 };
 
 const mapStateToProps = state => {
-  const { myAssets } = state;
-  return { myAssets };
+  const { myAssets, accountInfo } = state;
+  const { account } = accountInfo;
+  const { list: myAssetsList } = myAssets;
+  return { myAssets: myAssetsList, account };
```
pcowgill: Yeah, having it as an array in this component is a good choice

decentraland/helpers/decentraland.js
```diff
@@ -0,0 +1,156 @@
...
+
+  const estate = { type, id, name, img };
+  return estate;
+};
+
+const _generateParcelFromId = async (landContract, parcelId) => {
+  const id = `${parcelId}`;
+  const type = PARCEL;
+  const namePromise = landContract.tokenMetadata(id);
+  const coordsPromise = landContract.decodeTokenId(id);
```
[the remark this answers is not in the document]
marcelomorgado: No, not really. Read operations are async.

decentraland/screens/MyAssetsScreen.js
```diff
@@ -1,29 +1,63 @@
...
 export class MyAssetsScreen extends React.Component {
+  componentDidMount = async () => {
+    const { account, myAssets: fromState, setMyAssetsList } = this.props;
+    if (account) {
+      const { address } = account;
+      const fromBlockchain = await this._getAssetsFromBlockchain(address);
+      const shouldUpdate = !listsAreEqual(fromBlockchain, fromState);
+      if (shouldUpdate) {
+        setMyAssetsList(fromBlockchain);
+        storeMyAssets(fromBlockchain);
```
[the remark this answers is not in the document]
marcelomorgado: Yeah, I agree, it's related to the middleware discussion.

decentraland/screens/MyAssetsScreen.js
```diff
@@ -1,29 +1,63 @@
...
+      }
+    }
+  };
+
+  _getAssetsFromBlockchain = async address => {
+    const listOfPromises = await getAssetsOf(address);
+    const fromBlockchain = await Promise.all([...listOfPromises]);
```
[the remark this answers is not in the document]
marcelomorgado: Thanks

decentraland/helpers/decentraland.js
```diff
@@ -0,0 +1,156 @@
...
+  const listSize = 10;
+  for (let order of estatesForSale.slice(0, listSize)) {
+    let assetForSalePromise = _toAssetForSale(order);
+    assetsForSale.push(assetForSalePromise);
+  }
+
+  return assetsForSale;
+};
+
+const _toAssetForSale = async sellOrder => {
```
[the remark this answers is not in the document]
marcelomorgado: Oh, I've forgotten our design decision about it.
Even `_getAssetsOf` looks like that could be useful for other parts of the app, I've moved to `MyAssetsScreen.js` too since it's the only component that is using it. Agree?

decentraland/helpers/decentraland.js
marcelomorgado: Yes, done.

decentraland/screens/ListLandForSaleScreen.js
```diff
@@ -8,145 +8,29 @@ import {
...
-      return await this._prepareEstateForSale(estateContract, assetForSale);
-    } else if (isParcel) {
-      return await this._prepareParcelForSale(landContract, assetForSale);
-    } else {
-      throw new Error(`The asset should be a parcel of land or an estate.`);
+    try {
+      showInfo("Loading land for sale...");
+      const assetsForSale = await getAllAssetsForSale();
+      for (let promise of assetsForSale) {
+        const assetForSale = await promise;
```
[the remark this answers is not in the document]
marcelomorgado: Nice idea. Done with some style options for you.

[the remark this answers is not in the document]
pcowgill: Okay great, thanks

decentraland/helpers/decentraland.js
pcowgill: Thanks. Resolving

decentraland/helpers/decentraland.js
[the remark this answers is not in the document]
pcowgill: No worries. Yep, that makes sense to me. Thanks!

decentraland/screens/MyAssetsScreen.js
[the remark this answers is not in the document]
pcowgill: Resolving this

decentraland/screens/MyAssetsScreen.js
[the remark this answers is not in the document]
pcowgill: Yep, for sure. Maybe a comment above these two lines like ``// TODO: Maybe move me into a single function` would be sufficient for this PR?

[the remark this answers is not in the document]
pcowgill:
```suggestion
        // TODO: Maybe move these two set/store operations into a single function
        storeMyAssets(fromBlockchain);
```

decentraland/helpers/decentraland.js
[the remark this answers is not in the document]
pcowgill: In that case, do we want an `await` here?

decentraland/screens/ListLandForSaleScreen.js
```diff
@@ -20,17 +21,121 @@ export class ListLandForSaleScreen extends React.Component {
...
+      // Functional style
+      // const loadingAssetsOnScreen = assetsForSale.map(promise => {
+      //   let loadAssetOnScreen = async () => {
+      //     const assetForSale = await promise;
+      //     appendLandForSaleToList(assetForSale);
+      //   };
+      //   return loadAssetOnScreen();
+      // });
+
+      // Functional style with unnamed function
```
pcowgill: Note to self to weigh in on these options

decentraland/screens/ListLandForSaleScreen.js
```diff
@@ -20,17 +21,121 @@ export class ListLandForSaleScreen extends React.Component {
...
+    const listSize = 10;
+    for (let order of estatesForSale.slice(0, listSize)) {
+      let assetForSalePromise = this._toAssetForSale(order);
+      assetsForSale.push(assetForSalePromise);
+    }
+
+    return assetsForSale;
+  };
+
+  _toAssetForSale = async sellOrder => {
```
pcowgill: I almost wonder if `reformatAsset` might be a clearer name. When I see the to it makes me wonder what the from is, and I assume from the current name it doesn't involve and network I/O, but a name like reformat makes that very clear

decentraland/screens/MyAssetsScreen.js
pcowgill: Resolving this

decentraland/helpers/decentraland.js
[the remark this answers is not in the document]
pcowgill: Feel free to resolve this

decentraland/screens/ListLandForSaleScreen.js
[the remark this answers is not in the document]
pcowgill: We can move this discussion over to this thread https://github.com/tasitlabs/tasit/pull/259#discussion_r272656624

Resolving this

decentraland/screens/ListLandForSaleScreen.js
```diff
@@ -8,40 +8,72 @@ import {
...
     try {
       showInfo("Loading land for sale...");
-      await this._loadAssetsForSale();
+      const assetsForSale = await this._getAllAssetsForSale();
+
+      // Non-functional style
+      // const loadingAssetsOnScreen = [];
+      // for (let promise of assetsForSale) {
+      //   let loadAssetToScreen = async () => {
+      //     const assetForSale = await promise;
```
pcowgill: Wouldn't awaiting it here not add it to the array until this line completes, which would mean losing the parallelism?

decentraland/screens/ListLandForSaleScreen.js
```diff
@@ -8,40 +8,72 @@ import {
...
+      //     appendLandForSaleToList(assetForSale);
+      //   };
+      //   loadingAssetsOnScreen.push(loadAssetToScreen());
+      // }
+
+      // Functional style
+      // const loadingAssetsOnScreen = assetsForSale.map(promise => {
+      //   let loadAssetOnScreen = async () => {
+      //     const assetForSale = await promise;
+      //     appendLandForSaleToList(assetForSale);
```
pcowgill: I think I personally find this style (functional style with named function) the most readable, but others could certainly disagree for valid reasons. Do you have a strong preference?

decentraland/screens/ListLandForSaleScreen.js
pcowgill: I weighed in on these options, so I'm resolving this thread

decentraland/helpers/decentraland.js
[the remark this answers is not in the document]
marcelomorgado: In that case, we are getting both `name` and the `coords` parallelly using `Promise.all`

decentraland/helpers/decentraland.js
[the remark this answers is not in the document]
marcelomorgado: Resolving this.

decentraland/screens/MyAssetsScreen.js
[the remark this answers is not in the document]
marcelomorgado: Done.

decentraland/screens/ListLandForSaleScreen.js
marcelomorgado: I'm almost sure that won't. Following commented code how I understand it:

```javascript
// For each promise that will return an assetForSale...
const loadingAssetsOnScreen = assetsForSale.map(promise => {
  // Create a new function that returns a promise...
  let loadAssetOnScreen = async () => {
    const assetForSale = await promise;
    // to call the action function, the promise should be resolved
    appendLandForSaleToList(assetForSale);
  };
  // ... or in other words, warp a promise with another promise

   // returns a promise without awaiting it
  return loadAssetOnScreen();
});
// loadingAssetsOnScreen now has a list of promises (same as assetsForSale)
// and those promises will be waited by Promise.all()
```

decentraland/screens/ListLandForSaleScreen.js
marcelomorgado: No strong opinion, done.

decentraland/screens/ListLandForSaleScreen.js
marcelomorgado: I agree (that this name seems that it involves an I/O operation).
My intention here was something like `fromSellOrderToAssetForSale` because of I has assumed the name `sellOrder` to "raw data" from blockchain and `assetForSale` for the object used by the app as a way to avoid naming confusion.
Maybe the methods names from `DecentralandUtils` could be changed since its methods are returning "raw data".
Note: I'm used `raw data` to refs to data getted by the contract events, often `raw` data is related to hex data.

pcowgill: That all makes sense. Sometimes I even just use a `raw` prefix in the name but have the rest of the name be the same. So like `rawAssetForSale`

decentraland/screens/ListLandForSaleScreen.js
pcowgill: Okay, thanks! Resolving this.

decentraland/screens/ListLandForSaleScreen.js
pcowgill: Wait, but there's no for loop in your commented code. I agree that the map ones should work as you describe.

The line I commented had a for loop.

decentraland/screens/MyAssetsScreen.js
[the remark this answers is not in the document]
pcowgill: Thanks! Resolving this.

decentraland/screens/ListLandForSaleScreen.js
pcowgill: This discussion doesn't gate merging though, since we're going the map route

decentraland/helpers/decentraland.js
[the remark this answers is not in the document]
pcowgill: Ah right, because tokenMetadata doesn't ship the request yet, it just builds it. I forgot that. The use of promise in the name should have made that obvious to me. Thanks for explaining!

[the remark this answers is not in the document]
pcowgill: Resolving this

decentraland/screens/ListLandForSaleScreen.js
pcowgill: This change won't gate merging

decentraland/screens/ListLandForSaleScreen.js
marcelomorgado: > Wait, but there's no for loop in your commented code. I agree that the map ones should work as you describe.
> 
> The line I commented had a for loop.

Oh, you are right, my fault, I've thought that you are talking about the chosen style code. I think that the behavior will be the same on the `for` loop since the `await` is inside of the function and this function isn't being resolved inside of the loop. What do you think?

pcowgill: Yeah, I suppose because it's wrapped in that `let loadAssetToScreen = async () => {` it will work fine. I was thinking of a for loop where the first thing that happened was awaiting with the `async` scope outside of the toop. Resolving this.

decentraland/screens/ListLandForSaleScreen.js
marcelomorgado: Extracted to: https://github.com/tasitlabs/tasit/issues/273

pcowgill: Thanks! Resolving

PR #271, opened 2019-04-09
No large snapshots to error level

### LINKED ISSUE #268: No large snapshot - move rule to error level
We've setup the [`jest/no-large-snapshots`](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-large-snapshots.md) rule on our `eslint` and it's warning during the current test suite: 

> /.../tasit/decentraland/components/presentational/__snapshots__/**BuyLand**.test.js.snap
   87:1  warning  Expected Jest snapshot to be smaller than 50 lines but was 52 lines long  jest/no-large-snapshots
  141:1  warning  Expected Jest snapshot to be smaller than 50 lines but was 52 lines long  jest/no-large-snapshots
  195:1  warning  Expected Jest snapshot to be smaller than 50 lines but was 52 lines long  jest/no-large-snapshots
  249:1  warning  Expected Jest snapshot to be smaller than 50 lines but was 52 lines long  jest/no-large-snapshots
  303:1  warning  Expected Jest snapshot to be smaller than 50 lines but was 52 lines long  jest/no-large-snapshots
> 
> /.../tasit/decentraland/components/presentational/__snapshots__/**LandForSaleInfo**.test.js.snap
  3:1  warning  Expected Jest snapshot to be smaller than 50 lines but was 69 lines long  jest/no-large-snapshots

As I remember our motivation to turn this on is the avoid large components it still a valid motivation but these snapshots are large because we wrote many test cases for each one, there is no rule such as `no-large-snapshot-per-testcase`.
Here is a quote extracted from the rule description:

> As a best practice snapshots should be limited in size in order to be more manageable and reviewable.

Said that, should we change the way that we're testing our components keeping less test cases or increment the lines of each snapshot should have?

marcelomorgado:
### Auto-close the issue?

Closes #268 
 
### Types of changes

### Notes

Are you agree with that approach to keep the component in the same file if it is just used as an "helper" component?


pcowgill:
> Are you agree with that approach to keep the component in the same file if it is just used as an "helper" component?

Yep, that’s fine with me!

marcelomorgado:
> Any time we’re going to have a helper component, we should have one test that renders the parent component and one test that renders the helper component

Done.

pcowgill:
> > Any time we’re going to have a helper component, we should have one test that renders the parent component and one test that renders the helper component
> 
> Done.

Great, thanks!

decentraland/components/presentational/__snapshots__/AccountCreationProgress.test.js.snap
```diff
@@ -1,5 +1,103 @@
...
+  >
+    <Component
+      adjustsFontSizeToFit={true}
+      numberOfLines={2}
+    >
+      A waiting message.
+    </Component>
+  </Component>
+  <Component>
+    <LinkToBlockchain
```
pcowgill: Noting that we could have alternatively not rendered the component in the null action case as opposed to rendering it with a null action. I'm perfectly happy with the way we have it now - just mentioning it since I thought about it.

Issue #173, opened 2019-03-20
Show ephemeral account address and/or link to Etherscan transactions from within app


pcowgill:

pcowgill:
This is mostly for trying things out while testing the app and less of a "production-ready" feature

pcowgill:
@marcelomorgado One other benefit of this feature is that our alpha and beta testers trying the app would be able to confirm for themselves that indeed the back-end blockchain-related features are working as expected

marcelomorgado:
I'll work on that as my next issue.

marcelomorgado:
Where this link should be shown?

pcowgill:
> Where this link should be shown?

I'll be thinking about this shortly, and I'll let you know!

pcowgill:
@marcelomorgado I just added some rough thoughts about this in the `#design` channel on Slack, and once we come to a decision I'll post that here.

marcelomorgado:
After a discussion on Slack, we've decided these:

- Account setup transactions `etherscan` links should be shown on each step message;
- Move the user to `MyLandScreen` after purchase;
- On the `MyLandScreen`, show buy tx `etherscan` links below each parcel of land (using an eye icon);

marcelomorgado:
After purchase, if the user clicks on the `Land For Sale` menu item, he will be moved to the `BuyLandScreen` showing the asset already bought, It's a bug. The menu link 'Land for Sale' should move the user to the `LandForSaleListScreen`.

marcelomorgado:
@pcowgill commented about design:
> I’m thinking the eye for the link to each land tx can go to the lower right of the land image, right-aligned with the right side of the land image, with a little spacing between the land image and the eye

pcowgill:
> After purchase, if the user clicks on the `Land For Sale` menu item, he will be moved to the `BuyLandScreen` showing the asset already bought, It's a bug. The menu link 'Land for Sale' should move the user to the `LandForSaleListScreen`.

Should we split this out into a separate issue?

marcelomorgado:
> > After purchase, if the user clicks on the `Land For Sale` menu item, he will be moved to the `BuyLandScreen` showing the asset already bought, It's a bug. The menu link 'Land for Sale' should move the user to the `LandForSaleListScreen`.
> 
> Should we split this out into a separate issue?

Okay. Extracted to: https://github.com/tasitlabs/tasit/issues/267

marcelomorgado:
Updating the TODO list for this issue:

[...]

PR #298, opened 2019-04-16
Store account creation status

### LINKED ISSUE #296: Persist AccountInfo.creationActions state
- Persist `AccountInfo.creationActions` state (refs: https://github.com/tasitlabs/tasit/pull/293#issuecomment-483665892)
- ~~Tiny style change (refs: https://github.com/tasitlabs/tasit/pull/293#discussion_r275916140)~~

marcelomorgado:
### Issue link

#296 

### Auto-close the issue?

Closes #296 

### Types of changes

### DIFF
--- a/decentraland/App.js
+++ b/decentraland/App.js
@@ -6,13 +6,23 @@ import AppNavigator from "./AppNavigator";
 import { Provider } from "react-redux";
 import { createStore } from "redux";
 import decentralandApp from "./redux/reducers";
-import { setAccount, setMyAssetsList } from "./redux/actions";
+import {
+  setAccount,
+  setMyAssetsList,
+  setAccountCreationStatus,
+  setAccountCreationActions,
+} from "./redux/actions";
 import applyMiddleware from "./redux/middlewares";
 import { Action } from "tasit-sdk";
 const { ConfigLoader } = Action;
 import tasitSdkConfig from "./config/current";
 import { checkBlockchain, showFatalError } from "@helpers";
-import { retrieveEphemeralAccount, retrieveMyAssets } from "@helpers/storage";
+import {
+  retrieveAccount,
+  retrieveMyAssets,
+  retrieveAccountCreationStatus,
+  retrieveAccountCreationActions,
+} from "@helpers/storage";
 import { Ionicons } from "@expo/vector-icons";
 import { Root } from "native-base";
 
@@ -47,9 +57,15 @@ Is the config file correct?`;
     }
   }
 
-  async _loadAccount() {
-    const account = await retrieveEphemeralAccount();
-    if (account) store.dispatch(setAccount(account));
+  async _loadAccountInfo() {
+    const account = await retrieveAccount();
+    if (account) {
+      store.dispatch(setAccount(account));
+      const creationStatus = await retrieveAccountCreationStatus();
+      const creationActions = await retrieveAccountCreationActions();
+      store.dispatch(setAccountCreationStatus(creationStatus));
+      store.dispatch(setAccountCreationActions(creationActions));
+    }
   }
 
   async _loadMyAssets() {
@@ -81,10 +97,10 @@ Is the config file correct?`;
   _loadResourcesAsync = async () => {
     const setupSDK = this._setupTasitSDK();
     const loadFonts = this._loadFonts();
-    const loadAccount = this._loadAccount();
+    const loadAccountInfo = this._loadAccountInfo();
     const loadMyAssets = this._loadMyAssets();
 
-    return Promise.all([setupSDK, loadFonts, loadAccount, loadMyAssets]);
+    return Promise.all([setupSDK, loadFonts, loadAccountInfo, loadMyAssets]);
   };
 
   _handleLoadingError = error => {
--- a/decentraland/App.js
+++ b/decentraland/App.js
@@ -7,6 +7,7 @@ import { Provider } from "react-redux";
 import { createStore } from "redux";
 import decentralandApp from "./redux/reducers";
 import { setAccount, setMyAssetsList } from "./redux/actions";
+import applyMiddleware from "./redux/middlewares";
 import { Action } from "tasit-sdk";
 const { ConfigLoader } = Action;
 import tasitSdkConfig from "./config/current";
@@ -15,7 +16,7 @@ import { retrieveEphemeralAccount, retrieveMyAssets } from "@helpers/storage";
 import { Ionicons } from "@expo/vector-icons";
 import { Root } from "native-base";
 
-const store = createStore(decentralandApp);
+const store = createStore(decentralandApp, applyMiddleware);
 
 export default class App extends React.Component {
   state = {
--- a/decentraland/helpers/storage.js
+++ b/decentraland/helpers/storage.js
@@ -5,13 +5,42 @@ import { createFromPrivateKey } from "tasit-account/dist/testHelpers/helpers";
 // Storage keys
 const EPHEMERAL_ACCOUNT_PRIV_KEY = "EPHEMERAL_ACCOUNT_PRIV_KEY";
 const MY_ASSETS_LIST = "MY_ASSETS_LIST";
+const EPHEMERAL_ACCOUNT_CREATION_STATUS = "EPHEMERAL_ACCOUNT_CREATION_STATUS";
+const EPHEMERAL_ACCOUNT_CREATION_ACTIONS = "EPHEMERAL_ACCOUNT_CREATION_ACTIONS";
 
-export const storeEphemeralAccount = async account => {
+export const storeAccountCreationActions = async creationActions => {
+  const strCreationActions = _toString(creationActions);
+  await _storeData(
+    EPHEMERAL_ACCOUNT_CREATION_ACTIONS,
+    strCreationActions,
+    false
+  );
+};
+
+export const retrieveAccountCreationActions = async () => {
+  const strCreationActions = await _retrieveData(
+    EPHEMERAL_ACCOUNT_CREATION_ACTIONS,
+    false
+  );
+  let creationActions = _fromString(strCreationActions);
+  return creationActions;
+};
+
+export const storeAccountCreationStatus = async status => {
+  await _storeData(EPHEMERAL_ACCOUNT_CREATION_STATUS, status, false);
+};
+
+export const retrieveAccountCreationStatus = async () => {
+  const status = await _retrieveData(EPHEMERAL_ACCOUNT_CREATION_STATUS, false);
+  return status;
+};
+
+export const storeAccount = async account => {
   const { privateKey } = account;
   await _storeData(EPHEMERAL_ACCOUNT_PRIV_KEY, privateKey, true);
 };
 
-export const retrieveEphemeralAccount = async () => {
+export const retrieveAccount = async () => {
   let account = null;
 
   const privateKey = await _retrieveData(EPHEMERAL_ACCOUNT_PRIV_KEY, true);
@@ -80,8 +109,12 @@ const _retrieveData = async (key, securely) => {
 };
 
 export default {
-  storeEphemeralAccount,
-  retrieveEphemeralAccount,
+  storeAccount,
+  retrieveAccount,
   storeMyAssets,
   retrieveMyAssets,
+  storeAccountCreationStatus,
+  retrieveAccountCreationStatus,
+  storeAccountCreationActions,
+  retrieveAccountCreationActions,
 };
--- a/decentraland/redux/middlewares.js
+++ b/decentraland/redux/middlewares.js
@@ -7,16 +7,18 @@ import {
   REMOVE_MY_ASSET_FROM_LIST,
   SET_MY_ASSETS_LIST,
 } from "./actions";
-import { storeEphemeralAccount } from "@helpers/storage";
+import { storeEphemeralAccount, storeMyAssets } from "@helpers/storage";
 
 const storer = store => next => async action => {
   const { type } = action;
   const result = next(action);
-  //const nextState = store.getState();
+  const nextState = store.getState();
+  const { accountInfo, myAssets } = nextState;
+  const { account } = accountInfo;
+  const { list: myAssetsList } = myAssets;
 
   switch (type) {
     case SET_ACCOUNT: {
-      const { account } = action;
       await storeEphemeralAccount(account);
       break;
     }
@@ -27,12 +29,15 @@ const storer = store => next => async action => {
       break;
     }
     case ADD_TO_MY_ASSETS_LIST: {
+      await storeMyAssets(myAssetsList);
       break;
     }
     case REMOVE_MY_ASSET_FROM_LIST: {
+      await storeMyAssets(myAssetsList);
       break;
     }
     case SET_MY_ASSETS_LIST: {
+      await storeMyAssets(myAssetsList);
       break;
     }
   }
--- a/decentraland/redux/middlewares.js
+++ b/decentraland/redux/middlewares.js
@@ -3,29 +3,41 @@ import {
   SET_ACCOUNT,
   SET_ACCOUNT_CREATION_STATUS,
   UPDATE_ACTION_FOR_ACCOUNT_CREATION_STATUS,
+  SET_ACCOUNT_CREATION_ACTIONS,
   ADD_TO_MY_ASSETS_LIST,
   REMOVE_MY_ASSET_FROM_LIST,
   SET_MY_ASSETS_LIST,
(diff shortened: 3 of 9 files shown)


marcelomorgado:
Test failed because of lint (unnused var).

decentraland/helpers/storage.js
```diff
@@ -5,13 +5,42 @@ import { createFromPrivateKey } from "tasit-account/dist/testHelpers/helpers";
...
+    EPHEMERAL_ACCOUNT_CREATION_ACTIONS,
+    strCreationActions,
+    false
+  );
+};
+
+export const retrieveAccountCreationActions = async () => {
+  const strCreationActions = await _retrieveData(
+    EPHEMERAL_ACCOUNT_CREATION_ACTIONS,
+    false
```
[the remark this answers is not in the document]
pcowgill: (Not required to merge this, just a thought)

decentraland/App.js
```diff
@@ -46,9 +57,15 @@ Is the config file correct?`;
...
-    const account = await retrieveEphemeralAccount();
-    if (account) store.dispatch(setAccount(account));
+  async _loadAccountInfo() {
+    const account = await retrieveAccount();
+    if (account) {
+      store.dispatch(setAccount(account));
+      const creationStatus = await retrieveAccountCreationStatus();
+      const creationActions = await retrieveAccountCreationActions();
+      store.dispatch(setAccountCreationStatus(creationStatus));
+      store.dispatch(setAccountCreationActions(creationActions));
```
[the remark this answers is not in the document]
pcowgill: (Not required to merge this, just a thought)

decentraland/App.js
```diff
@@ -47,9 +57,15 @@ Is the config file correct?`;
...
 
-  async _loadAccount() {
-    const account = await retrieveEphemeralAccount();
-    if (account) store.dispatch(setAccount(account));
+  async _loadAccountInfo() {
+    const account = await retrieveAccount();
+    if (account) {
+      store.dispatch(setAccount(account));
+      const creationStatus = await retrieveAccountCreationStatus();
+      const creationActions = await retrieveAccountCreationActions();
```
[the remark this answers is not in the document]
pcowgill: I supposed maybe the function can be stringified and un-stringified?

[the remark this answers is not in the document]
pcowgill: We’ll want to test the `getLink` function or whatever it’s called to confirm

decentraland/redux/middlewares.js
```diff
@@ -0,0 +1,60 @@
...
+      break;
+    }
+    case REMOVE_MY_ASSET_FROM_LIST: {
+      await storeMyAssets(myAssetsList);
+      break;
+    }
+    case SET_MY_ASSETS_LIST: {
+      await storeMyAssets(myAssetsList);
+      break;
+    }
```
pcowgill: (Comment on somewhat random line) Do we need any error handling code in this middleware?

decentraland/redux/middlewares.js
```diff
@@ -0,0 +1,60 @@
...
+  SET_MY_ASSETS_LIST,
+} from "./actions";
+import {
+  storeAccount,
+  storeAccountCreationStatus,
+  storeAccountCreationActions,
+  storeMyAssets,
+} from "@helpers/storage";
+
+const storer = store => next => async action => {
```
pcowgill: async middleware for a sync reducer is fine then, right?

decentraland/redux/middlewares.js
```diff
@@ -0,0 +1,60 @@
...
+  const { type } = action;
+  const result = next(action);
+  const nextState = store.getState();
```
pcowgill: Did you decide that as long as it’s in middleware this isn’t a code smell?

decentraland/App.js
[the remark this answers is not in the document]
marcelomorgado: Since the `MyAccountScreen` isn't linking to blockchain for now, it'll work.
But I'm already aware of this issue and I'm addressing it now when I'm working on https://github.com/tasitlabs/tasit/issues/173 I'm thinking to store the `actionId` instead of `action` object.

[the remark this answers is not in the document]
pcowgill: Okay, sounds good. And then we would do a `new action(...)` built from the actionId? Or use a separate link building function that goes straight from the id?

decentraland/redux/middlewares.js
marcelomorgado: Yes, a middleware can return a `Promise` (Ref: https://redux.js.org/advanced/middleware#seven-examples > vanillaPromise)
About promises error handler: 

```javascript
**
 * Lets you dispatch promises in addition to actions.
 * If the promise is resolved, its result will be dispatched as an action.
 * The promise is returned from `dispatch` so the caller may handle rejection.
 */
```

decentraland/redux/middlewares.js
marcelomorgado: At first glance, the size doesn't seem too long for me, but we can split into smaller middlewares or to helpers functions if you prefer.

pcowgill: No, it seems fine to me! I thought maybe you originally were concerned about the use of getStore for this purpose more so than any size-related concerns. Either way, feel free to resolve this. Thanks!

decentraland/App.js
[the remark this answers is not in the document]
marcelomorgado: As far as I've checked, change `LinkToBlockchain` component to receive an `actionId` will be enough. If the `action` object having a `buildLink` action we'll still need to resolve it and store (instead of the actionId only) because it'll be an `async` function.

decentraland/redux/middlewares.js
pcowgill: Okay, cool.

So will we await the dispatch of these actions from now on?

On the error handling topic, I'll note that in their example they use the promise before the action itself, and here we're doing the opposite. Which I think is okay - I'm pretty sure our returning the value from the async function after dispatching the sync action will work too.

decentraland/redux/middlewares.js
marcelomorgado: Oh, got it now! :smile: 
Yes, since the middleware provides the store object I think that isn't a code smell.
Resolving this.

decentraland/App.js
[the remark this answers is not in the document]
pcowgill: I agree that we can add another option for `LinkToBlockchain` to work that way in addition to working with a full URL.

If we started keeping the actionid (a.k.a. tx hash) as part of the action object, `buildLink` wouldn't necessarily need to be an async function.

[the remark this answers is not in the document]
pcowgill: What do you mean by "If the `action` object having a `buildLink` action"? I had trouble parsing that part of your comment.

decentraland/redux/middlewares.js
pcowgill: Okay, great!

decentraland/redux/middlewares.js
marcelomorgado: Hmmm
On the example, I've understood that since the middleware is returning a `Promise` we can write our middleware function as `async` too. And the example's action is a promise, we can do that from now but none of our actions are async. Make sense?

decentraland/App.js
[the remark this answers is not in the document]
marcelomorgado: Oh, I've meant an `action.buildLink` action to encapsulate the `txHash` same as `action.getId` I've thought that you made reference to this.

decentraland/redux/middlewares.js
marcelomorgado: - For action that was dispatched manually (those from `App` component), we can catch it and show an error on UI.
- Not sure if it will be possible to do that (show on UI) from actions dispatched by redux.

pcowgill: `Not sure if it will be possible to do that (show on UI) from actions dispatched by redux.`

What's an example of an action dispatched by redux that "we" don't trigger?

decentraland/redux/middlewares.js
pcowgill: > On the example, I've understood that since the middleware is returning a `Promise` we can write our middleware function as `async` too.

> And the example's action is a promise, we can do that from now but none of our actions are async. Make sense?

I would say that the example uses a promise in the middleware, but the action itself is not a promise. Note that the comment above the vanilla example says, "If the promise is resolved, its result will be dispatched as an action."

decentraland/App.js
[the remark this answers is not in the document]
pcowgill: Hmm, the way you're saying "`action.buildLink action`" with the word action used twice is still confusing me.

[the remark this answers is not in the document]
pcowgill: I did make a reference to the `action.buildLink` method/function though, yep.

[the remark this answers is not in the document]
pcowgill: > I agree that we can add another option for `LinkToBlockchain` to work that way in addition to working with a full URL.
> 
> If we started keeping the actionid (a.k.a. tx hash) as part of the action object, `buildLink` wouldn't necessarily need to be an async function.

Thoughts on this?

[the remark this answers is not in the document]
marcelomorgado: Yes, I agree. I think that we could store the `actionId` from now and change to the URL in the future.

decentraland/redux/middlewares.js
marcelomorgado: Maybe I'm wrong, but I've undestood from the docs that 

`store.dispatch(setAccountCreationStatus(creationStatus));` (App.js) will throw error from middleware but `setAccountCreationStatus(creationStatus)` (EthereumSignUpScreen.js) won't.

decentraland/redux/middlewares.js
marcelomorgado: Here is the full example:

```javascript
/**
 * Lets you dispatch promises in addition to actions.
 * If the promise is resolved, its result will be dispatched as an action.
 * The promise is returned from `dispatch` so the caller may handle rejection.
 */
const vanillaPromise = store => next => action => {
  if (typeof action.then !== 'function') {
    return next(action)
  }

  return Promise.resolve(action).then(store.dispatch)
}
```

Seems to me that the action is async since it is being resolved by `Promise`. Is it right?

decentraland/redux/middlewares.js
pcowgill: > Maybe I'm wrong, but I've undestood from the docs that
> 
> `store.dispatch(setAccountCreationStatus(creationStatus));` (App.js) will throw error from middleware but `setAccountCreationStatus(creationStatus)` (EthereumSignUpScreen.js) won't.

I understand now. That makes sense to me. I think you're right about this. So for cases where we can't throw, we still might want to log the error from within the middleware.

decentraland/App.js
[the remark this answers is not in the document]
pcowgill: Feel free to resolve this if you like.

decentraland/helpers/storage.js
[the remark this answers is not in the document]
pcowgill: Feel free to resolve this if you like.

decentraland/App.js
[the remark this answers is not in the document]
pcowgill: Works for me!

decentraland/redux/middlewares.js
pcowgill: I don't think so, specifically because of the comment I quoted, `"If the promise is resolved, its result will be dispatched as an action."`.

The action is passed through the promise, but the "`redux action`" itself doesn't happen until store.dispatch is called. So the action is still sync. I guess this is sort of a semantics question more than anything, though.

decentraland/App.js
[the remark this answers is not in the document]
pcowgill: So making this change will gate merging this PR then, right?

decentraland/redux/middlewares.js
marcelomorgado: Oh, I see. 
This discussion is helping me to understand redux a little better.
Please, correct if I'm wrong, but what is dispatched is the result of an action function, for example `{ type: REMOVE_LAND_FOR_SALE, landForSale }` so, we can write an async action function that returns this object and do the same check as the example (if the action function is or an async or not). Make sense?

decentraland/redux/middlewares.js
marcelomorgado: OK, should I add some `console.log` commands thru the middleware?

pcowgill: I think so, if that works for you. I forget, do we have tooling in place to strip console.log for production builds yet?

decentraland/App.js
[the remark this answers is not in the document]
marcelomorgado: We can merge this PR, since the account creation links are shown only from the `BuyLandScreen`. I'm using the `actionId` on a WIP branch right now, and I think that we'll be done in a few hours.

decentraland/App.js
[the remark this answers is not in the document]
marcelomorgado: I've read again and not sure if I got it, could you explain using other words?

decentraland/helpers/storage.js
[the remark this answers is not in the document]
marcelomorgado: Resolving this.

decentraland/redux/middlewares.js
marcelomorgado: We have it on babel config but have to be done manually.

decentraland/redux/middlewares.js
pcowgill: I'm using this definition of action. "Actions are payloads of information that send data from your application to your store."

> Please, correct if I'm wrong, but what is dispatched is the result of an action function

So I would say it like this: The dispatch of the action can be delayed until receiving the result of an async function with middleware written one way. Alternatively, an action can be dispatched by a wrapper function that also triggers an async function after the action has been dispatched (and I think that's what we're doing as it's written here).

> so, we can write an async action function that returns this object and do the same check as the example (if the action function is or an async or not). Make sense?

I don't think we can write an async function that handles actions, but I wouldn't call it an "action function".

"Action creators are exactly that—functions that create actions. It's easy to conflate the terms “action” and “action creator”, so do your best to use the proper term." Source: https://redux.js.org/basics/actions

We definitely can check if the action creator function is async or not.

decentraland/redux/middlewares.js
pcowgill: What has to be done manually?

decentraland/App.js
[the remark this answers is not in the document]
pcowgill: The middleware is already intercepting certain actions and persisting them. If at the top of the middleware, there were an array of actions that required persistence, we could export that array for the middleware and import it here. Then, we could loop through that array to decide on app load which pieces of data to retrieve and then hydrate redux with.

decentraland/App.js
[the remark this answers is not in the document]
pcowgill: As long as the function on the action object can be stringified and un-stringified without creating errors - I assume you've tested this. That works for me!

decentraland/redux/middlewares.js
marcelomorgado: Oh.. great thanks! I was misunderstanding `action` x `action creator`. It's all clear now.

decentraland/redux/middlewares.js
marcelomorgado: Uncomment `plugins: ["transform-remove-console"],` from `babel.config.js`.

pcowgill: Ah ok. Any reason not to have it always enabled?

decentraland/redux/middlewares.js
pcowgill: Great!

pcowgill: The way redux uses action is actually quite similar to the way we define it. A thing that can be sent to the data store right away or later (redux vs. Ethereum blockchain) which is a state mutation.

decentraland/redux/middlewares.js
marcelomorgado: If we want to have the app ready for a develop env, make sense make that way otherwise no debug message/error will be showed on the console.

pcowgill: Oh right. So what we really want to do is write a little bit of logic that enables or disables this based on env.

pcowgill: Or more likely write our own tiny logger helper that does that same thing without using the babel feature

marcelomorgado: https://github.com/tasitlabs/tasit/pull/323

pcowgill: Awesome, thanks! So if we want to go back and add a console.log here in the middleware, we can now

Issue #291, opened 2019-04-15
Improve look of "My account" screen


pcowgill:
@marcelomorgado I think if we're going to have a header, we should probably move the progress bar up directly below the header, touching it.

Also, I think the individual progress contributors should be aligned towards the top of the screen rather than towards the bottom.

Think this should gate merging https://github.com/tasitlabs/tasit/pull/287 into `develop`? I'd say it's fine to merge into `develop` as-is, as long as we address this before running `npm run publish` or `npm run build:[ios/android]`

![Simulator Screen Shot - iPhone Xʀ - 2019-04-15 at 18 36 16](https://user-images.githubusercontent.com/2731712/56172094-882ad480-5fad-11e9-8d32-c2993f5c7319.png)

pcowgill:
This is how it looks when it's all complete, as of right now

![Simulator Screen Shot - iPhone Xʀ - 2019-04-15 at 18 40 31](https://user-images.githubusercontent.com/2731712/56172224-0dae8480-5fae-11e9-8de7-ddf26fecb6ba.png)

pcowgill:
I can see now how the wireframe was ambiguous, and what we have so far is certainly a reasonable interpretation of what is in the design. But seeing the real thing, I think a few small improvements are needed.

https://www.figma.com/file/0vJIc39s2tbVJ8VzMihyI7pI/My-profile?node-id=0%3A1&redirected=1

pcowgill:
The individual progress items could also be vertically centered if we had less spacing between each item

pcowgill:
FYI I went ahead and merged #287 into `develop`. Let’s not build or publish from `develop` until we tweak the layout a bit, though. Normally that could be a dangerous practice, but the layout changes here are very minor.

PR #301, opened 2019-04-17
Additional tweaks to my account style

pcowgill:
### Issue link

https://github.com/tasitlabs/tasit/issues/291

##### Other (please describe below if you select "Other")
Styling

### Notes

Simulator Screen Shot - iPhone Xʀ - 2019-04-16 at 21 02 59

### DIFF
--- a/decentraland/components/presentational/MyAccountCreationStatusItem.js
+++ b/decentraland/components/presentational/MyAccountCreationStatusItem.js
@@ -18,13 +18,15 @@ export default function MyAccountCreationStatusItem({ name, status }) {
   );
 }
 
+const ICON_SIZE = 30;
+const FONT_SIZE = (ICON_SIZE * 7) / 10;
+
 function renderIcon(status) {
-  const ICON_SIZE = 20;
   if (status === ActionStatus.DONE) {
     return (
       <Ionicons
         name="md-checkmark"
-        size={ICON_SIZE}
+        size={(ICON_SIZE * 3) / 5}
         style={styles.actionStatusIcon}
       />
     );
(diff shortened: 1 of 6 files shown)


decentraland/components/presentational/MyAccountCreationStatusItem.js
```diff
@@ -18,13 +18,15 @@ export default function MyAccountCreationStatusItem({ name, status }) {
...
 }
 
+const ICON_SIZE = 30;
+const FONT_SIZE = (ICON_SIZE * 7) / 10;
+
 function renderIcon(status) {
-  const ICON_SIZE = 20;
   if (status === ActionStatus.DONE) {
     return (
       <Ionicons
```
marcelomorgado: Make sense to use icons from [`native-base`](https://docs.nativebase.io/Components.html#icon-def-headref)?

pcowgill: No strong preference. Happy to make that change. Will the icons look more native on each platform? Are they guaranteed to have a fixed width unlike these icons?

marcelomorgado: Seems that `native-base` `Icon` component uses `Ionicicons` by default. About size:

> In case if you want to include icon with custom color, size etc then that should go into style.

pcowgill: Ah ok. Then maybe I’ll stick with the Expo-bundled Ionicons since we’re more likely to move away from NativeBase some day than we are from Expo. Noted about styling, thanks

pcowgill: Resolving this

PR #315, opened 2019-04-18
Loading data from blockchain for old accounts

### LINKED ISSUE #306: Restore account creation state from blockchain (for old accounts)
Since until now, we were only storing the `ephemeralAccount` on local storage the `MyAccountScreen` is showing a buggy '25%' state (25% the progress for an account without funds). It will not happen for new accounts since from now we are storing entire account create state. For old accounts we need to check the blockchain to restore the state.

Note: `isOldAccount = retrieveAccount() !== null && retrieveAccountCreationStatus() === null && retrieveAccountCreationActions() === null`

marcelomorgado:
### Issue link

#306 

### Auto-close the issue?

Closes #306 

### Types of changes


decentraland/App.js
```diff
@@ -61,11 +65,23 @@ Is the config file correct?`;
...
+      let creationActions = await retrieveAccountCreationActions();
+
+      const isOldAccount =
+        account !== null && creationStatus === null && creationActions === null;
+
+      if (isOldAccount) {
+        ({
+          creationStatus,
+          creationActions,
+        } = await restoreCreationStateOfAccountFromBlockchain(account));
```
pcowgill: We should test if this method works by checking out an old commit and then running this code

decentraland/App.js
```diff
@@ -97,12 +113,20 @@ Is the config file correct?`;
 
   // More about AppLoading: https://docs.expo.io/versions/latest/sdk/app-loading/
   _loadResourcesAsync = async () => {
-    const setupSDK = this._setupTasitSDK();
     const loadFonts = this._loadFonts();
-    const loadAccountInfo = this._loadAccountInfo();
     const loadMyAssets = this._loadMyAssets();
 
-    return Promise.all([setupSDK, loadFonts, loadAccountInfo, loadMyAssets]);
+    // Note: _loadAccountInfo needs that TasitSDK is ready to use
```
pcowgill: Good call

decentraland/helpers/index.js
```diff
@@ -178,6 +188,46 @@ export const openURL = async url => {
...
+  let creationActions = {};
+
+  const accountCreated = account !== null;
+  const fundedWithEthers = `${ethersBalance}` !== `${ZERO}`;
+  const fundedWithMana = `${manaBalance}` !== `${ZERO}`;
+  const marketplaceWasApproved = `${allowance}` != `${ZERO}`;
+
+  if (accountCreated) {
+    creationStatus = FUNDING_WITH_ETH;
+    creationActions[FUNDING_WITH_ETH] = null;
```
pcowgill: Hmm, I think I personally would have modified the logic on the `My account` screen to use balances if it's an old account rather than putting fake actions in storage

decentraland/helpers/index.js
```diff
@@ -178,6 +188,46 @@ export const openURL = async url => {
...
+  }
+
+  if (fundedWithEthers) {
+    creationStatus = FUNDING_WITH_MANA_AND_APPROVING_MARKETPLACE;
+    creationActions[APPROVING_MARKETPLACE] = null;
+    creationActions[FUNDING_WITH_MANA] = null;
+  }
+
```
pcowgill: Maybe handle the case here where either mana fanding or marketplace approve worked but not the other, given that it would just a few more lines, even if it's unlikely?

pcowgill: This would be to set creationStatus, and we'd probably want an `else if` to handle this

pcowgill:
Nice work!!

decentraland/helpers/index.js
pcowgill: Partially because that screen will want access to that data even for new accounts pretty soon

decentraland/App.js
marcelomorgado: I've made this test here setting `creationStatus` and `creationActions` to `null`

decentraland/helpers/index.js
marcelomorgado: I've added the `APPROVING_MARKETPLACE` and `FUNDING_WITH_MANA` status, I haven't used the `else if`, I've tried to reproduce the same flow from `signup` using independent `ifs`

decentraland/App.js
pcowgill: Yep, I agree it ought to work in theory - just thought maybe we should be paranoid since we have some users out there now

decentraland/App.js

decentraland/helpers/index.js
pcowgill: (This won’t gate merging this PR)

decentraland/App.js
marcelomorgado: Ok, I got it

decentraland/helpers/index.js
pcowgill: Yeah true, the ordering of ifs can achieve the same effect as using else if

decentraland/helpers/index.js
marcelomorgado: I've just finished a refactoring on this component, I'll work to get account progress from blockchain now.

pcowgill: Oh ok, nice!!

decentraland/helpers/index.js
```diff
@@ -221,6 +221,14 @@ export const restoreCreationStateOfAccountFromBlockchain = async account => {
     creationActions[FUNDING_WITH_MANA] = null;
   }
 
+  if (fundedWithMana || !marketplaceWasApproved) {
```
pcowgill: Hmm, this `OR` seems dangerous to me. Maybe we should use an `AND` instead? Is there a specific case you want to catch with the `OR`?

decentraland/helpers/index.js
marcelomorgado: Ok, great. Extract to: https://github.com/tasitlabs/tasit/issues/318
So, I'll do the test for an old account right now.

decentraland/helpers/testHelpers.js
```diff
@@ -41,18 +41,18 @@ export const anAction = {
   getId: async () => "0x1234567890123456789012345678901234567890",
 };
 
-export const accountCreationActions = [
+export const accountCreationSteps = [
   {
     name: "Account created",
-    status: ActionStatus.DONE,
+    status: DONE,
```
pcowgill: Ooh yeah, this is much nicer

decentraland/screens/MyAccountScreen.js
```diff
@@ -1,53 +1,73 @@
...
+    const { creationActions, account } = accountInfo;
+    const isAccountCreated = account !== null;
+
+    const creationSteps = [];
+
+    creationSteps.push({
       name: "Account created",
-      action: null,
-      status: !accountInfo.account ? ActionStatus.MISSING : ActionStatus.DONE,
+      creationStatus: null,
```
pcowgill: What does this get used for?

pcowgill: Or put another way, why push things onto `creationSteps` in two spots? (This happens a few lines up and then lower down in this file.)

decentraland/screens/MyAccountScreen.js
```diff
@@ -1,53 +1,73 @@
...
-      );
-      if (isDone && hasDefinedActionPercentage) {
-        percentage += AccountCreationActions[accountAction.action].percentage;
-      }
-    });
+  _getPercentage(creationSteps) {
+    const percentage = creationSteps
+      .filter(step => step.status === DONE)
+      .map(step => step.percentage)
+      .reduce((total, stepPercentage) => total + stepPercentage, 0);
```
pcowgill: I haven't used reduce much yet. It's interesting that it has access to `total` in this way.

decentraland/helpers/index.js
pcowgill: Awesome, thanks!

decentraland/helpers/index.js
marcelomorgado: Oh, logic error. Fixed. Thanks!

decentraland/screens/MyAccountScreen.js
marcelomorgado: Yes, neither do I, seems a good way to do accumulations.

pcowgill: Yep, cool. Resolving this.

decentraland/helpers/index.js
pcowgill: Thanks. Resolving this.

decentraland/helpers/testHelpers.js
pcowgill: Resolving this

decentraland/helpers/index.js
pcowgill: Resolving this

decentraland/helpers/index.js
pcowgill: Resolving this

decentraland/screens/MyAccountScreen.js
marcelomorgado: During account generation, no data is being saved on `accountInfo.creationActions` and the loop is looking into this object to determine if it was done or not.
Note: It seems odd to me too, I've just refactored the code keeping the pre-existent logic.

pcowgill: Okay, thanks. You haven't pushed up that change yet, right?

marcelomorgado: Just done.

decentraland/App.js
marcelomorgado: . Clear app data
. Checkout from commit `ca55a44be00e74a5fa79927949a980fc0af11d02`
. Start app
. MyAccountScreen shows 0%
. Account created
. MyAccountScreen shows 100%
. Restart app
. MyAccountScreen shows 25%
. Checkout from branch `feature/25-percent`
. Start app
. MyAccountScreen shows 25% :x: 

Trying to figured out

pcowgill: Hmm, weird. Okay, thanks. Maybe add some temporary log statements throughout the code base?

decentraland/screens/MyAccountScreen.js
```diff
@@ -27,46 +33,49 @@ const StepsWithAction = [
...
-      creationStatus: null,
-      status: isAccountCreated ? DONE : MISSING,
-      percentage: 0.25,
-    });
+  if (creationStatus === GENERATING_ACCOUNT) {
+    const isAccountCreated = account !== null;
+    return isAccountCreated;
+  } else if (statusWithAction.includes(creationStatus)) {
+    const hasAnAction = creationActions[creationStatus];
+    return hasAnAction;
```
pcowgill: It's a little weird that sometimes the function returns a boolean and sometimes not.

marcelomorgado: Both are treated as boolean outside, just make it's explicit.

pcowgill: Resolved this

decentraland/screens/MyAccountScreen.js
pcowgill: Resolving this

decentraland/App.js
marcelomorgado: Just fixed.

decentraland/screens/MyAccountScreen.js
marcelomorgado: The fix was related to this function.

pcowgill: Ah ok interesting.

decentraland/App.js
pcowgill: Woo! Great job.

pcowgill: Should we include a test that catches this case or are we all set as-is?

marcelomorgado: Good idea

pcowgill: I tried the steps above out to be double-paranoid, and I can confirm it's showing 100% for me at the end as it should

marcelomorgado: Done on PR https://github.com/tasitlabs/tasit/pull/325

PR #325, opened 2019-04-19
MyAccountScreen tests

marcelomorgado:
### Issue link

https://github.com/tasitlabs/tasit/pull/315#discussion_r277054288

### Types of changes

### Notes

Here is possible to verify that "too optimistic" behavior that I've said on Slack.
When an account is funded with ethers both mana and approval steps run in parallel updating the `creationActions` state object with `FUNDING_WITH_MANA` and `APPROVING_MARKETPLACE` making the `MyAccountScreen` shows the progress as 100%.

### DIFF
--- a/decentraland/screens/MyAccountScreen.test.js
+++ b/decentraland/screens/MyAccountScreen.test.js
@@ -37,7 +37,7 @@ describe("MyAccountScreen", () => {
       ).toMatchSnapshot();
     });
 
-    it("funding with ethers with null action (used when action is unknowed)", async () => {
+    it("funding with ethers with null action (used when action is unknown)", async () => {
       expect(
         shallow(
           <MyAccountScreen
--- a/decentraland/screens/MyAccountScreen.test.js
+++ b/decentraland/screens/MyAccountScreen.test.js
@@ -1,11 +1,15 @@
 import React from "react";
 import { shallow } from "enzyme";
 import { MyAccountScreen } from "./MyAccountScreen";
-import AccountCreationStatus from "@constants/AccountCreationStatus";
+import {
+  FUNDING_WITH_ETH,
+  FUNDING_WITH_MANA,
+  APPROVING_MARKETPLACE,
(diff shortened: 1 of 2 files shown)


decentraland/screens/MyAccountScreen.test.js
```diff
@@ -18,14 +22,46 @@ describe("MyAccountScreen", () => {
...
-                [AccountCreationStatus.FUNDING_WITH_ETH]: {},
+                [FUNDING_WITH_ETH]: {},
+              },
+            }}
+          />
+        )
+      ).toMatchSnapshot();
+    });
+
+    it("funding with ethers with null action (used when action is unknowed)", async () => {
```
pcowgill:
```suggestion
    it("funding with ethers with null action (used when action is unknown)", async () => {
```

decentraland/screens/__snapshots__/MyAccountScreen.test.js.snap
```diff
@@ -34,7 +34,75 @@ exports[`MyAccountScreen renders the component with account 1`] = `
...
+        "creationStatus": "FUNDING_WITH_ETH",
+        "name": "Funded with ETH",
+        "percentage": 0.25,
+        "status": "done",
+      },
+      Object {
+        "creationStatus": "FUNDING_WITH_MANA",
+        "name": "Funded with MANA tokens",
+        "percentage": 0.25,
+        "status": "done",
```
pcowgill: Ah ok, so this isn't a bug though based on the way we've designed it. Both funding with mana and approving marketplace have been initiated, and until we store pending state, it's the best we can do

pcowgill:
@marcelomorgado Tests failed since I renamed that test but didn't update the snapshot (and remove the one with the old name)

marcelomorgado:
> @marcelomorgado Tests failed since I renamed that test but didn't update the snapshot (and remove the one with the old name)

Fixed.

PR #331, opened 2019-04-22
Prep for forcing account removal after uninstall

### LINKED ISSUE #295: Force account removal after app uninstallation
iOS is keeping the user's account after app uninstallation (Related to: https://github.com/tasitlabs/tasit/issues/289#issuecomment-483706543).
This behavior happens with `SecureStore` only. To solve it, alike `isFirstAppUse` should be persisted on `AsyncStore` and if the condition ` isFirstAppUse && retrieveEphemeralAccount() != null` the stored account should be removed.

marcelomorgado:
### Issue link

https://github.com/tasitlabs/tasit/pull/298#discussion_r276225655
#295 

### Auto-close the issue?

Closes #295

### Types of changes

### Notes

The way that it is, all accounts will be removed because none of them have the param `isFirstUse` stored yet.
Maybe we can introduce `storeIsFirstUse(false);` only for now and check this var (and remove account) later? Any better idea?

### DIFF
--- a/decentraland/App.js
+++ b/decentraland/App.js
@@ -69,7 +69,8 @@ Is the config file correct?`;
     // On iOS environment, Secure Store data remains even after app uninstallation
     const isFirstUse = await retrieveIsFirstUse();
     if (isFirstUse) {
-      await storeAccount(null);
+      // TODO: Uncomment this when 'isFirstUse' flag is being set widely
+      //await storeAccount(null);
       await storeIsFirstUse(false);
     }
 
--- a/decentraland/App.js
+++ b/decentraland/App.js
@@ -26,7 +26,6 @@ import {
   retrieveMyAssets,
   retrieveAccountCreationStatus,
   retrieveAccountCreationActions,
-  storeAccount,
   storeIsFirstUse,
   retrieveIsFirstUse,
 } from "@helpers/storage";
--- a/decentraland/App.js
+++ b/decentraland/App.js
@@ -26,6 +26,9 @@ import {
   retrieveMyAssets,
   retrieveAccountCreationStatus,
   retrieveAccountCreationActions,
+  storeAccount,
+  storeIsFirstUse,
+  retrieveIsFirstUse,
 } from "@helpers/storage";
 import { Ionicons } from "@expo/vector-icons";
 import { Root } from "native-base";
@@ -62,6 +65,14 @@ Is the config file correct?`;
   }
 
   async _loadAccountInfo() {
+    // Note: Forcing account removal after app uninstall
+    // On iOS environment, Secure Store data remains even after app uninstallation
+    const isFirstUse = await retrieveIsFirstUse();
+    if (isFirstUse) {
+      await storeAccount(null);
+      await storeIsFirstUse(false);
+    }
+
     const account = await retrieveAccount();
     if (account) {
       store.dispatch(setAccount(account));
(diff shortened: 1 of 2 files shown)


pcowgill:
> Maybe we can introduce storeIsFirstUse(false); only for now and check this var (and remove account) later? Any better idea?

This works for me, with a TODO to remove it soon.

pcowgill:
> > Maybe we can introduce storeIsFirstUse(false); only for now and check this var (and remove account) later? Any better idea?
> 
> This works for me, with a TODO to remove it soon.

We could get fancy with checking app version from the manifest to determine behavior, but that's too much overhead for now

decentraland/helpers/storage.js
```diff
@@ -7,6 +7,19 @@ const EPHEMERAL_ACCOUNT_PRIV_KEY = "EPHEMERAL_ACCOUNT_PRIV_KEY";
 const MY_ASSETS_LIST = "MY_ASSETS_LIST";
 const EPHEMERAL_ACCOUNT_CREATION_STATUS = "EPHEMERAL_ACCOUNT_CREATION_STATUS";
 const EPHEMERAL_ACCOUNT_CREATION_ACTIONS = "EPHEMERAL_ACCOUNT_CREATION_ACTIONS";
+const APP_IS_FIRST_USE = "APP_IS_FIRST_USE";
+
+export const storeIsFirstUse = async isFirstUse => {
+  const strIsFirstUse = _toString(isFirstUse);
+  await _storeData(APP_IS_FIRST_USE, strIsFirstUse, false);
```
pcowgill: Maybe `IS_FIRST_APP_USE` would make more sense?

pcowgill: Resolving this

pcowgill: Or wait, I should unresolve this

pcowgill: Haha sorry, I was looking at the old commits instead of the new ones. I can resolve this after all.

decentraland/App.js
```diff
@@ -69,7 +69,8 @@ Is the config file correct?`;
     // On iOS environment, Secure Store data remains even after app uninstallation
     const isFirstUse = await retrieveIsFirstUse();
     if (isFirstUse) {
-      await storeAccount(null);
+      // TODO: Uncomment this when 'isFirstUse' flag is being set widely
+      //await storeAccount(null);
```
pcowgill: Oh, rather than this I was picturing still removing the account if `isFirstUse`, but in the new app version I would set `isFirstUse` to `false` for all users on first app load and then remove that later

decentraland/helpers/storage.js
```diff
@@ -48,7 +48,7 @@ export const retrieveAccountCreationStatus = async () => {
 };
 
 export const storeAccount = async account => {
-  const { privateKey } = account;
+  const privateKey = account !== null ? account.privateKey : null;
```
pcowgill: Ah yes, good catch

decentraland/App.js
pcowgill: I'll wait for your thoughts on this before merging, but I'm happy with leaving it as you've done it if you prefer

pcowgill:
I think we can remove `WIP` from the name of this PR

decentraland/App.js
pcowgill: Never mind, I'll merge this without waiting on this

pcowgill:
I renamed the PR to say "prep for" because only once we un-comment that line in a future PR will the feature be done. For now it's preparation for that.

decentraland/App.js
marcelomorgado: Hmmm, if that way works for you, it seems ok for me.
I think that sometime soon when we uncomment `await storeAccount(null);` we'll only remove accounts after uninstallation (the exception will be users that we'll pass all that timeframe without using the app).

pcowgill: Yep, agreed. Sounds good. Resolving this.

pcowgill: And noted about that caveat - I had thought about that but it’s okay.

Issue #197, opened 2019-03-26
Decentraland: How is the best way to store loading states?


marcelomorgado:
We are encapsulating both `account` and `assetsList` loading states inside of their objects (on redux reducers).
Should we keep that way or keep them on the same higher level?
This question was originated from this thread: https://github.com/tasitlabs/tasit/pull/194#discussion_r269168560

marcelomorgado:
I've made a quick read about React Suspense and seems with that manage loading flags won't be necessary.

Refs:
https://blog.logrocket.com/async-rendering-in-react-with-suspense-5d0eaac886c8
https://hackernoon.com/magic-of-react-suspense-with-concurrent-react-and-react-lazy-api-e32dc5f30ed1

pcowgill:
Thanks for looking into this. I'll take a look sometime soon!

PR #333, opened 2019-04-22
MyLand refresh/navigation fix

### LINKED ISSUE #267: After purchase navigation
After purchase, user is being moved to `ListLandForSaleScreen`, to improve the UI flow, that navigation should be change to redirect the user to `MyAssetsScreen`.
We've made a tryout on that and faced some issues that need to me addressed:

- After purchase, if the user clicks on the `Land For Sale` menu item, he will be moved to the `BuyLandScreen` showing the asset already bought, It's a bug. The menu link 'Land for Sale' should 
move the user to the `LandForSaleListScreen`.

- Since we are getting the list of users' land from blockchain during `MyAssetsScreen.componentDidMount` what is happening is:
On the first time of the app: User signs up, buy land and before the purchase process finished on the blockchain, he is moved to the `MyAssetsScreen` that will get no assets from blockchain and the list will be empty.
On the second use, the list will be loaded with the land bought on first use and so on...
The expected behavior only will happen if the user access the `MyAssetsScreen` before purcharse any asset.

marcelomorgado:
### Issue link

#267 

### Auto-close the issue?

Closes #267 

### Types of changes

### Notes

Could you tell more about your last comment on the issue?

> I think the pattern to follow here is that there's a redux state update for each network request when request is initiated, when it succeeds, and when it fails due to timeout or error.


pcowgill:
> > I think the pattern to follow here is that there's a redux state update for each network request when request is initiated, when it succeeds, and when it fails due to timeout or error.

> Could you tell more about your last comment on the issue?

Sure! Each type of network request - A) fetching my land; B) fetching land for sale; C) fetching account funding plus approval should have a name for that network request, and then when it it starts we save to redux that network request with name `BLA_BLA` is now in one of the enum list of options for network request status, between 1) loading, 2) succeeded, 3) failed.

The relevant one here would be the my land request.

Although going too far down this path might not be the best use of time because of the React Suspense discussion here https://github.com/tasitlabs/tasit/issues/197

Does that clarify it, or should I go into more detail?

marcelomorgado:
> > > I think the pattern to follow here is that there's a redux state update for each network request when request is initiated, when it succeeds, and when it fails due to timeout or error.
> 
> > Could you tell more about your last comment on the issue?
> 
> Sure! Each type of network request - A) fetching my land; B) fetching land for sale; C) fetching account funding plus approval should have a name for that network request, and then when it it starts we save to redux that network request with name `BLA_BLA` is now in one of the enum list of options for network request status, between 1) loading, 2) succeeded, 3) failed.
> 
> The relevant one here would be the my land request.
> 
> Although going too far down this path might not be the best use of time because of the React Suspense discussion here #197
> 
> Does that clarify it, or should I go into more detail?

Oh, I see, something like this:

```javascript
{ type: 'FETCH_POSTS_REQUEST' }
{ type: 'FETCH_POSTS_FAILURE', error: 'Oops' }
{ type: 'FETCH_POSTS_SUCCESS', response: { ... } }
```

Extracted from redux doc.

pcowgill:
> Oh, I see, something like this:
> 
> ```js
> { type: 'FETCH_POSTS_REQUEST' }
> { type: 'FETCH_POSTS_FAILURE', error: 'Oops' }
> { type: 'FETCH_POSTS_SUCCESS', response: { ... } }
> ```
> 
> Extracted from redux doc.

Yep, exactly

pcowgill:
@marcelomorgado I know there were a couple issues being tracked in issue #267 . Have some of them since become obsolete? Just checking whether this PR indeed closes all of them.

decentraland/helpers/index.js
```diff
@@ -171,6 +177,13 @@ export const listsAreEqual = (first, second) => {
   );
 };
 
+// Update any list having id as key field
```
pcowgill:
```suggestion
// Update item from any list of objects having id as key field
```

marcelomorgado:
> @marcelomorgado I know there were a couple issues being tracked in issue #267 . Have some of them since become obsolete? Just checking whether this PR indeed closes all of them.

This PR is addressing:
- Fix: Refreshing from blockchain was removed on-process purchase lands;
- Fix: After redirect user to `MyLandScreen` the menu link `Land For Sale` jumped the user to Buy Screen of already bought land.
- UX improvement: Redirect the user to `MyLandScreen` after clicking on `Buy` button (only possible after fixing both bugs above).

decentraland/redux/reducers.js
```diff
@@ -83,13 +85,26 @@ function assetsForSale(state = { list: [], loadingInProgress: true }, action) {
...
+    myAssetAndStatus,
+    itemOrList,
+  } = action;
+
   switch (type) {
-    case ADD_TO_MY_ASSETS_LIST:
+    case PREPEND_TO_MY_ASSETS_LIST:
       return { ...state, list: [myAsset, ...state.list] };
-    case REMOVE_MY_ASSET_FROM_LIST: {
+    case APPEND_TO_MY_ASSETS_LIST: {
```
pcowgill: @marcelomorgado Out of curiosity, do you think this `myAssets` reducer function is getting a little large to reason about? I'm wondering whether we can split it out into a few functions that each handle fewer different type cases.

(Comment made on somewhat arbitrary line)

decentraland/screens/BuyLandScreen.js
```diff
@@ -64,13 +69,15 @@ export class BuyLandScreen extends React.Component {
       // that catches the safeExecuteOrder successful event.
       await action.waitForNonceToUpdate();
 
+      updateMyAssetStatus(assetId, BOUGHT);
```
pcowgill: I wonder about generalizing this term bought to something that could happen for any write tx. Like pending? And meta-tx might have state `BROADCAST` but not yet `PENDING`

decentraland/constants/MyAssetStatus.js
```diff
@@ -0,0 +1,7 @@
+export const BUYING = "BUYING";
+export const BOUGHT = "BOUGHT";
+
+export default {
+  BUYING,
+  BOUGHT,
+};
```
pcowgill: We'll definitely want a failed one here too

decentraland/screens/BuyLandScreen.js
```diff
@@ -80,9 +87,11 @@ export class BuyLandScreen extends React.Component {
 
     // Optimistic UI update
     removeLandForSale(landForSale);
-    addToMyAssetsList(asset);
+    prependToMyAssetsList({ ...asset, status: BUYING });
 
-    navigation.navigate("ListLandForSaleScreen");
+    // Back to top of current Stack before navigate
+    navigation.dispatch(StackActions.popToTop());
```
pcowgill: Interesting 👍

decentraland/screens/BuyLandScreen.test.js
```diff
@@ -28,10 +29,11 @@ describe("BuyLandScreen", () => {
             myAssets={myAssets}
             selectedLandToBuy={estateForSale}
             removeLandForSale={removeLandForSale}
-            addToMyAssetsList={addToMyAssetsList}
-            removeMyAssetFromList={removeMyAssetFromList}
+            prependToMyAssetsList={prependToMyAssetsList}
+            removeFromMyAssetsList={removeFromMyAssetsList}
             prependLandForSaleToList={prependLandForSaleToList}
             setActionIdForMyAsset={setActionIdForMyAsset}
+            updateMyAssetStatus={updateMyAssetStatus}
```
pcowgill: It's starting to feel odd the number of functions we're needing to pass through to this component

decentraland/screens/MyAssetsScreen.js
```diff
@@ -1,25 +1,36 @@
...
+    const {
+      account,
+      myAssets: assetsFromState,
+      removeFromMyAssetsList,
+      appendToMyAssetsList,
+    } = this.props;
     if (account) {
       const { address } = account;
+
+      // Note: If an stored asset hasn't a status, assuming 'bought'.
```
pcowgill:
```suggestion
      // Note: If a stored asset doesn't have a status, assuming 'bought'.
```

decentraland/screens/MyAssetsScreen.js
```diff
@@ -1,25 +1,36 @@
...
       const assetsFromBlockchain = await this._getAssetsFromBlockchain(address);
-      const shouldUpdate = !listsAreEqual(
-        assetsFromBlockchain,
-        assetsFromState
-      );
+      const shouldUpdate = !listsAreEqual(assetsFromBlockchain, boughtAssets);
       if (shouldUpdate) {
-        setMyAssetsList(assetsFromBlockchain);
+        removeFromMyAssetsList(boughtAssets);
+        appendToMyAssetsList(assetsFromBlockchain);
```
pcowgill: If an asset was locally there but ends up being removed as a result of this, I think we should log something at a minimum, and eventually add some UI indication that something unexpected happened

pcowgill:
> > @marcelomorgado I know there were a couple issues being tracked in issue #267 . Have some of them since become obsolete? Just checking whether this PR indeed closes all of them.
> 
> This PR is addressing:
> 
>     * Fix: Refreshing from blockchain was removed on-process purchase lands;
> 
>     * Fix: After redirect user to `MyLandScreen` the menu link `Land For Sale` was jumped the user to Buy Screen of already bought land.
> 
>     * UX improvement: Redirect the user to `MyLandScreen` after clicking on `Buy` button (only possible after fixing both bugs above).

Awesome, this is a big improvement!!

decentraland/redux/reducers.js
marcelomorgado: I agree. I've followed some tips from https://redux.js.org/recipes/structuring-reducers/refactoring-reducer-example seems better now? We can do the same for all reducers or even extract a file for each reducer.

decentraland/screens/BuyLandScreen.js
marcelomorgado: I'm using `BOUGHT` as a successful status. Maybe make more sense when we were using event listeners (i.e.: confs > 7)?

decentraland/redux/reducers.js
pcowgill: Okay, sounds good, thanks! I'll take a look.

decentraland/constants/MyAssetStatus.js
marcelomorgado: I agree, but for now, we aren't tracking error events, should I add to the Constant anyway?

decentraland/redux/reducers.js
```diff
@@ -85,51 +85,69 @@ function assetsForSale(state = { list: [], loadingInProgress: true }, action) {
...
+      return setActionIdForMyAsset(state, action);
+    case UPDATE_MY_ASSET_STATUS:
+      return updateMyAssetStatus(state, action);
     default:
       return state;
   }
 }
 
+const prependToMyAssetsList = (state, action) => {
+  const { myAsset } = action;
```
pcowgill: Is there a version of this where the action can just be the relevant piece of data for the reducer that will handle it? It seems redundant to also have to unpack a key/value that we already could have guessed based on the action name

pcowgill: Hmm, I think I take that back. Because the name of the action is also part of the action object

decentraland/redux/reducers.js
pcowgill: Yep, I like the change to this reducer. Resolving this!

decentraland/constants/MyAssetStatus.js
pcowgill: Maybe just a TODO in that file rather than actually adding it.

pcowgill: Even if it's not an error from the contract we can track, we might use it if the phone is offline and the request itself fails to be made

decentraland/screens/BuyLandScreen.js
pcowgill: Yep. But I'm saying since we might want to use this pattern for the MANA funding, for instance, where `BOUGHT` wouldn't describe the success state well.

That said, keeping an action-specific name for success might be better - I'm not sure.

pcowgill: But yeah, we can wait until we're using event listeners more to formalize a standard around how to handle this stuff

decentraland/screens/BuyLandScreen.test.js
marcelomorgado: Yes, what is the best way to handle with this? Merge some actions?
Example:
```javascript
/// from
    removeLandForSale(landForSale);
    prependToMyAssetsList({ ...asset, status: BUYING });
// to
    moveAssetFromForSaleToMyAssets(landForSale, { ...asset, status: BUYING });
```

I'm not sure how the best way having an action changing more than one reducer (and if it's desirable). I'll take a look.

decentraland/redux/reducers.js
marcelomorgado: I've found this on the [redux docs](https://redux.js.org/recipes/structuring-reducers/refactoring-reducer-example#combining-reducers-by-slice):
```javascript
function createReducer(initialState, handlers) {
  return function reducer(state = initialState, action) {
    if (handlers.hasOwnProperty(action.type)) {
      return handlers[action.type](state, action)
    } else {
      return state
    }
  }
}

const todosReducer = createReducer([], {
  ADD_TODO: addTodo,
  TOGGLE_TODO: toggleTodo,
  EDIT_TODO: editTodo
})
```

decentraland/screens/BuyLandScreen.test.js
pcowgill: Hmm, let's see. I'm not sure why we would need:

`myAssets`

on the buy land screen.

Then I think this component can have one function called "process buy" for a piece of land.

Then that function up in whichever component / screen it's originally defined can handle splitting out into multiple actions for all the relevant state that's updated (removeLandForSale, prependToMyAssetsList, and setActionIdForMyAsset, and updateMyAssetStatus).

And for `removeFromMyAssetsList`, I'm not sure from first principles why that would be used when buying land anyway.

There's still only one reducer affected by any given action, but multiple actions are dispatched.

pcowgill: Hmm, only some of what I just said makes sense. I forgot this was the top level screen. I guess there are some functions that maybe should be defined outside of the scope of a given screen though, so they can react even if the user navigates away in the meantime?

decentraland/redux/reducers.js
pcowgill: Ah yeah, exactly. Up to you whether that feels like a useful change to make or not.

decentraland/screens/BuyLandScreen.js
pcowgill: Resolving this

decentraland/screens/BuyLandScreen.test.js
pcowgill: So many of those functions are wrapped into the _buy function. I just didn't realize that the test would need them all defined in this way. Does it need to be passed in like this in the test for it to work?

marcelomorgado: We can remove action functions from Proptype requirements, if we do so, it won't be needed.

decentraland/redux/reducers.js
marcelomorgado: Done.

decentraland/constants/MyAssetStatus.js
marcelomorgado: I was thinking about that and these statuses were modeled to be used on the assets on the `MyLandScreen`.
If a purchase failed it'll be added to `LandForSaleScreen` again (it could have status but we aren't using it for now only on `MyLandScreen`). 
Maybe will be better having a status for the action instead of an action for the asset?

decentraland/screens/BuyLandScreen.test.js
pcowgill: Hmm yeah, what value do we get out of `mapDispatchToProps`? It's possible we get some - I'm just not used to that approach.

decentraland/constants/MyAssetStatus.js
pcowgill: > Yeah, I think we want a data structure with all relevant actions not related to account funding that closely mirrors that data structure.

So both the my land screen and the list land for sale screen would use the state of that purchase actions data structure to reflect what is shown on its screen.

decentraland/redux/reducers.js
pcowgill: Great, thanks. Resolving this.

decentraland/screens/BuyLandScreen.test.js
marcelomorgado: When we want to test some action behavior we still are able to pass it as props even if it wasn't on the `proptype` requirements. The only `pros` that I see is to have more strict component props definition. But I think worth more having less boilerplate.

decentraland/screens/MyAssetsScreen.js
marcelomorgado: Since it's related to `Toast` make sense extract that?

pcowgill: Hmm, I think that may sound good but I'm not totally sure what you mean.

decentraland/screens/BuyLandScreen.test.js
pcowgill: Yeah, if we were using them in the test it would make sense - and maybe that's a change we should make in the future and add an issue for. But for now the enzyme test isn't using them in any way https://github.com/tasitlabs/tasit/blob/766f5a4e8386e76ac4b7ec1e4a701e8bd67adfad/decentraland/screens/BuyLandScreen.test.js

pcowgill: I think we can have a strict component props definition for the top-level buy function but not the functions that function uses. If we ever want to test the subfuntctions with mocks, we can compile a mock version of the buy function from mocked subfunctions. Does that make sense?

decentraland/constants/MyAssetStatus.js
marcelomorgado: Ok.
We have `assetsForSale` and `myAssets` reducers.
An `assetForSale` stores the `asset` and other sell order info like `price` and `seller`.
On `myAssets` we store only the assets (Since a user could have an asset without buying from a marketplace).
I'm thinking about how will be the best way to store the `status` field. 
We are already storing the `actionId` (txhash) on the `asset` object, I'm thinking to move the status to here and change following your thoughts:
Before: `asset = { id, name, img, actionId, status }`
After: `asset = { id, name, img, action: { id, status } }`
Action status could be: `PENDING`, `FAILED`, `CONFIRMED` or `SUCCESSFULL`.
`CONFIRMED` or `SUCCESSFULL` could be splited soon to distinguish between 1-conf and 7-conf.
And that could be reused for all actions on the app. Make sense?

decentraland/screens/BuyLandScreen.test.js
marcelomorgado: I'm not sure since all functions from redux are passed as component prop separately.

decentraland/screens/MyAssetsScreen.js
marcelomorgado: > add some UI indication that something unexpected happened

I thought to use `Toast` component for that, but it isn't working properly right now.

decentraland/redux/reducers.js
```diff
@@ -21,90 +21,105 @@ import { removeFromList, updateListItem, toListIfNot } from "@helpers";
...
-      return { ...state, list };
-    }
-    case SET_LOADING_ASSETS_FOR_SALE_IN_PROGRESS:
-      return { ...state, loadingInProgress };
-    default:
-      return state;
-  }
-}
+const selectedLandToBuy = createReducer(null, {
+  [SELECT_LAND_TO_BUY]: selectLandToBuy,
```
pcowgill: The past tense for one name (selected) - the wrapped reducer - and present tense internal function (select) is a little confusing. Maybe let's pick one or the other?

decentraland/redux/reducers.js
```diff
@@ -148,6 +163,21 @@ const updateMyAssetStatus = (state, action) => {
...
+    [APPEND_TO_MY_ASSETS_LIST]: appendToMyAssetsList,
+    [REMOVE_FROM_MY_ASSETS_LIST]: removeFromMyAssetsList,
+    [SET_MY_ASSETS_LIST]: setMyAssetsList,
+    [SET_ACTION_ID_FOR_MY_ASSET]: setActionIdForMyAsset,
+    [UPDATE_MY_ASSET_STATUS]: updateMyAssetStatus,
+  }
+);
+
+//
+// All all reducers
```
pcowgill: All all -> all

decentraland/redux/reducers.js
marcelomorgado: `selectedLandToBuy` is the reducer/state that stores the selected land from user
`selectLandToBuy` is the action to select
Maybe change `selectedLandToBuy` to `landToBuy`?

decentraland/screens/BuyLandScreen.test.js
marcelomorgado: Oh wait, I've just realize that we are validating proptypes on `eslint` setup and if we'll just remove the action frunctions from we'll need to change the eslint config. I'll take a look if has a recommendation for the situation like ours.

decentraland/redux/reducers.js
pcowgill: Thanks for addressing this. Resolving.

decentraland/redux/reducers.js
pcowgill: Yeah, I like that suggestion. Thanks!

decentraland/screens/MyAssetsScreen.js
pcowgill: Oh yeah. Plus I think we're maybe using Toasts a little bit too much? They work well as a placeholder for future UI elements, though. Maybe just log something for now and add a TODO comment in the code to reflect it in the UI?

decentraland/screens/BuyLandScreen.test.js
pcowgill: > I'm not sure since all functions from redux are passed as component prop separately.

Do they need to be?

decentraland/constants/MyAssetStatus.js
pcowgill: SUCCESSFULL -> SUCCESSFUL

pcowgill: > `CONFIRMED` or `SUCCESSFUL` could be splited soon to distinguish between 1-conf and 7-conf.

Love this idea!

pcowgill: > After: `asset = { id, name, img, action: { id, status } }`

This looks good to me!

pcowgill: > On `myAssets` we store only the assets (Since a user could have an asset without buying from a marketplace).

Makes sense. In this case the actionId would be for the transfer tx instead.

pcowgill: Oh wait, there's one thing I'm suggesting that maybe isn't clear.

Let me know if this is too big of a refactor for now.

I think there should be a separate data structure in redux called `userActions`. It's just an object of all the actions the user has taken, keyed maybe by actionId?

(1) myAssets doesn't contain anything about the actions other than maybe actionId as a pointer. Either that OR
(2) `userActions` has a pointer to the assetId for each action and myAssts has nothing about actions.

I think I like option 2 better.

The myAssets screen uses both `userActions` and `myAssets` state from redux to figure out what to show.

The listLandForSale screen uses both `userActions` and `assetsForSale` to figure out what to show.

This is a little more complex at first glance, but this way we're not updating action status for a given assets in two spots, in both `myAssets` and `assetsForSale`

decentraland/screens/BuyLandScreen.test.js
pcowgill: > Oh wait, I've just realize that we are validating proptypes on `eslint` setup and if we'll just remove the action functions from we'll need to change the eslint config. I'll take a look if has a recommendation for the situation like ours.

This is a naive question, but can we stop mapping dispatch to props?

pcowgill: Otherwise, maybe this is "code smell" that this really is just one "action", and our reducers + datastore have been designed in too complicated of a way and need to find a way to handle this single action.

decentraland/constants/MyAssetStatus.js
marcelomorgado: Nice!
I like option 2 too because we could have more than one action per asset (e.g.: many attempts before success) tracked too.

decentraland/screens/BuyLandScreen.test.js
marcelomorgado: Hmmmm maybe:
```javascript
const mapDispatchToProps = {
  removeLandForSale,
  prependLandForSaleToList,
  prependToMyAssetsList,
  removeFromMyAssetsList,
  setActionIdForMyAsset,
  updateMyAssetStatus,
};

// could be ?
const mapDispatchToProps = {
   buyFunctions: {
     removeLandForSale,
     prependLandForSaleToList,
     prependToMyAssetsList,
     removeFromMyAssetsList,
     setActionIdForMyAsset,
     updateMyAssetStatus, 
  }
};
```

decentraland/screens/MyAssetsScreen.js
marcelomorgado: OK

decentraland/constants/MyAssetStatus.js
pcowgill: Yep, good point!

decentraland/screens/MyAssetsScreen.js
marcelomorgado: Done. Should we log more detailed info (e.g.: which assetIds aren't on the blockchain)?

pcowgill: Sure!

decentraland/screens/BuyLandScreen.test.js
pcowgill: What do you think about multiple reducers reacting to a single action instead?

pcowgill: There's an egghead lesson about it, so it must not be too much of an antipattern

https://egghead.io/lessons/react-respond-to-a-single-redux-action-in-multiple-reducers

marcelomorgado: Hmmm, I'll take a look on that.

decentraland/redux/reducers.js
marcelomorgado: Done.

pcowgill: Thanks. Resolving this.

decentraland/screens/MyAssetsScreen.js
```diff
@@ -1,25 +1,40 @@
...
-      const shouldUpdate = !listsAreEqual(
-        assetsFromBlockchain,
-        assetsFromState
-      );
+      const shouldUpdate = !listsAreEqual(assetsFromBlockchain, boughtAssets);
       if (shouldUpdate) {
-        setMyAssetsList(assetsFromBlockchain);
+        // TODO: Add some UI indication that something unexpected happened
+        logWarn(
+          `Accounts' assets from blockchain aren't the same as the stored on the app.`
```
pcowgill:
```suggestion
          `Account's assets from blockchain aren't the same as the ones stored on the app.`
```

decentraland/constants/MyAssetStatus.js
pcowgill: This change should generalize well to other apps too, so that’s good

decentraland/screens/MyAssetsScreen.js
marcelomorgado: Done.

decentraland/screens/BuyLandScreen.js
marcelomorgado: I think that this thread is related to the below one?

pcowgill: Yep, i think so. So we'll use `PENDING`, `FAILED`, `CONFIRMED` or `SUCCESSFUL` for this instead of `BOUGHT`, etc.?

decentraland/constants/MyAssetStatus.js
pcowgill: > Action status could be: `PENDING`, `FAILED`, `CONFIRMED` or `SUCCESSFUL`

And meta-tx or signed messages for contract-based accounts might have state `BROADCAST` but not yet `PENDING`

decentraland/screens/BuyLandScreen.js
marcelomorgado: Yes. I'm thinking to use `SUCCESSFUL` because it is a term less related to the blockchain. Make sense?

decentraland/screens/MyAssetsScreen.js
pcowgill: Thanks! Resolving this

decentraland/screens/BuyLandScreen.js
pcowgill: Yep, sounds good. We'll use SUCCESSFUL for the current bought state after nonce updates.

And then whenever we track a single confirmation vs. 7, successful will be for 7 confirmations and something like `PROBABLY_SUCCESSFUL` will be for 1 confirmation

decentraland/screens/MyAssetsScreen.js
```diff
@@ -30,15 +30,33 @@ export class MyAssetsScreen extends React.Component {
...
+    const removedIds = fromStateIds.filter(
+      id => !fromBlockchainIds.includes(id)
+    );
+
+    logWarn(
+      `Account's assets from blockchain aren't the same as the ones stored on the app.`
+    );
+
+    if (addedIds.length > 0)
+      logWarn(`Some assets added to MyLandScreen. IDs: [${addedIds}]`);
```
pcowgill: This could probably be a `logInfo` rather than a `logWarn`. Either that or promote the removal one up from `logWarn` to `logError`

decentraland/constants/MyAssetStatus.js
marcelomorgado: Done.

pcowgill: Awesome, thanks!! Resolving this

decentraland/screens/MyAssetsScreen.js
marcelomorgado: Done.

decentraland/App.js
```diff
@@ -11,6 +11,7 @@ import {
   setMyAssetsList,
   setAccountCreationStatus,
   setAccountCreationActions,
+  addUserAction,
```
pcowgill: We can keep this as-is, but I first read it as "the action of adding a user"

pcowgill: (And I know it's my fault for calling it `userAction` haha)

decentraland/App.js
```diff
@@ -99,6 +102,21 @@ Is the config file correct?`;
 
   async _loadMyAssets() {
     const myAssets = await retrieveMyAssets();
+    const userActions = await retrieveUserActions();
+
+    // Handling with actions stored before
+    // the introduction of userAction state (<= 0.0.17)
```
pcowgill: Good call adding this

decentraland/components/presentational/MyAssetsList.js
```diff
@@ -11,27 +11,35 @@ import LargeText from "@presentational/LargeText";
 // https://medium.com/groww-engineering/stateless-component-vs-pure-component-d2af88a1200b
 export default class MyAssetsList extends React.PureComponent {
   render() {
-    const { myAssetsList, renderItem } = this.props;
-    const { length: listAmount } = myAssetsList;
+    const { myAssets, userActions, renderItem } = this.props;
+    const { length: listAmount } = myAssets;
```
pcowgill: I forget, is length explicitly defined or is this a newer ECMAScript feature?

decentraland/constants/UserActionStatus.js
```diff
@@ -0,0 +1,17 @@
+// To be used for meta-tx
+export const BROADCASTED = "BROADCASTED";
+
+export const PENDING = "PENDING";
+export const FAILED = "FAILED";
+
+// Note: We are using only 'SUCCESSFUL' for now
+// When we'll tracking confirmations that will be used for >= 7
```
pcowgill: we'll -> we're

decentraland/constants/UserActionStatus.js
```diff
@@ -0,0 +1,17 @@
+// To be used for meta-tx
```
pcowgill: Also for contract-based accounts with signed messages that will be bundled into tx with enough signatures (which is a meta-tx in a sense but I'd argue not what people normally mean by meta-tx

decentraland/helpers/storage.js
```diff
@@ -8,6 +8,19 @@ const MY_ASSETS_LIST = "MY_ASSETS_LIST";
...
+
+export const storeUserActions = async userActions => {
+  const strUserActions = _toString(userActions);
+  await _storeData(USER_ACTIONS, strUserActions, false);
+};
+
+export const retrieveUserActions = async () => {
+  const strUserActions = await _retrieveData(USER_ACTIONS);
+  const userActions = _fromString(strUserActions);
+  if (userActions === null) return [];
```
pcowgill: So we're storing `userActions` as an array rather than an object keyed by some id?

decentraland/screens/MyAssetsScreen.js
```diff
@@ -1,42 +1,76 @@
...
+        addUserAction(actionsFromBlockchain);
       }
     }
   };
 
+  _getAssetsAndActionsFromBlockchain = async address => {
+    const assetsFromBlockchain = await this._getAssetsFromBlockchain(address);
+
+    const actionsFromBlockchain = assetsFromBlockchain.map(asset => {
+      const { actionId, id: assetId } = asset;
```
pcowgill: It's interesting that the `_getAssetsFromBlockchain` function has `actionId` available. I wouldn't have guessed that. I'll need to take another look at that function.

decentraland/screens/MyAssetsScreen.js
```diff
@@ -1,42 +1,76 @@
...
+      const status = SUCCESSFUL;
+      return { actionId, assetId, status };
+    });
+
+    assetsFromBlockchain.forEach(asset => delete asset.actionId);
```
pcowgill: Is this temporary, or will we always need to do this?

decentraland/helpers/storage.js
pcowgill: As long as userActions and the assets list looping through both won't be a problem. If it ever becomes one, we can change one from an array to an object with a key as an "index"

pcowgill: Are the "account info" actions an array or an object?

decentraland/screens/MyAssetsScreen.js
```diff
@@ -80,15 +80,15 @@ export class MyAssetsScreen extends React.Component {
...
       `Account's assets from blockchain aren't the same as the ones stored on the app.`
     );
 
     if (addedIds.length > 0)
-      logWarn(`Some assets added to MyLandScreen. IDs: [${addedIds}]`);
+      logInfo(`Some assets added to MyLandScreen. IDs: [${addedIds}]`);
 
     if (removedIds.length > 0)
-      logWarn(`Some assets removed from MyLandScreen. IDs: [${removedIds}]`);
+      logInfo(`Some assets removed from MyLandScreen. IDs: [${removedIds}]`);
```
pcowgill: Wouldn't we still want this one to be a warn, though?

decentraland/screens/MyAssetsScreen.js
pcowgill: Awesome, thanks. One remaining thing on this thread over here:

https://github.com/tasitlabs/tasit/pull/333/commits/edb45201c90a443512e3e92005ba9b63bad15494#r277888035

decentraland/screens/MyAssetsScreen.js
pcowgill: (Not gating merging this PR)

decentraland/screens/MyAssetsScreen.js
pcowgill: (Not gating merging this PR)

decentraland/helpers/storage.js
pcowgill: (Not gating merging this PR)

decentraland/constants/UserActionStatus.js
pcowgill:
```suggestion
// To be used for meta-tx
// Also for contract-based accounts with signed messages
// that will be bundled into tx with enough signatures 
```

decentraland/constants/UserActionStatus.js
pcowgill:
```suggestion
// When we're tracking confirmations that will be used for >= 7
```

decentraland/components/presentational/MyAssetsList.js
pcowgill: (Not gating merging this PR)

decentraland/App.js
pcowgill: (Not gating merging this PR)

decentraland/App.js
pcowgill: (Not gating merging this PR)

decentraland/screens/BuyLandScreen.js
pcowgill: Resolving this

decentraland/screens/BuyLandScreen.test.js
pcowgill: I'm thinking we can make a new issue for this and merge this PR without making this change.

decentraland/screens/MyAssetsScreen.js
pcowgill: At first I thought this might as well come in this PR, but given that it's the end of the day, I'll merge the PR as-is and we can circle back to this.

decentraland/components/presentational/MyAssetsList.js
marcelomorgado: Every array has the length property

decentraland/helpers/storage.js
marcelomorgado: It is an array. But I think it should be an object.

decentraland/screens/MyAssetsScreen.js
marcelomorgado: It comes from `DecentralandUtils`, each `sell order` has its `transactionHash` prop that is being renamed to `actionId` inside of `_getAssetsFromBlockchain` function.

decentraland/screens/MyAssetsScreen.js
marcelomorgado: Since `sellOrders` from `DecentralandUtils` have `transactionHash`/`actionId` and we want to keep this info only on `userActions` state, I think it could be needed.

decentraland/screens/MyAssetsScreen.js
marcelomorgado: > Wouldn't we still want this one to be a warn, though?

Yes, make sense.

decentraland/screens/BuyLandScreen.test.js
marcelomorgado: Extracted to: https://github.com/tasitlabs/tasit/issues/341

decentraland/screens/MyAssetsScreen.js
marcelomorgado: Okay, I'll resolve this one since the other thread is tracking it.

decentraland/screens/MyAssetsScreen.js
marcelomorgado: Extracted to: https://github.com/tasitlabs/tasit/issues/342

decentraland/helpers/storage.js
marcelomorgado: Extracted to: https://github.com/tasitlabs/tasit/issues/342

decentraland/App.js
marcelomorgado: Extracted to: https://github.com/tasitlabs/tasit/issues/342

decentraland/components/presentational/MyAssetsList.js
pcowgill: Ahh yeah of course, this is basically the same as doing `myAssets.length`, and I remember that that's possible haha

pcowgill: Resolving this

decentraland/helpers/storage.js
pcowgill: Cool, thanks. Resolving this.

decentraland/screens/MyAssetsScreen.js
pcowgill: Ah right, I think that makes sense - one question first thought:

When you say `sell order` here do you mean the listing of the item for sale in the marketplace or the execution of the order by the buyer?

decentraland/screens/MyAssetsScreen.js
pcowgill: Thanks for the explanation. Is that the most compact/convenient way for the utils to return this information given how we'll use it going forward?

decentraland/screens/MyAssetsScreen.js
pcowgill: Thanks! Resolving this

decentraland/screens/BuyLandScreen.test.js
pcowgill: Thanks. Resolving this.

decentraland/screens/MyAssetsScreen.js
pcowgill: Good call. Thanks.

decentraland/App.js
pcowgill: Thanks! Resolving this.

decentraland/App.js

decentraland/screens/MyAssetsScreen.js
marcelomorgado: the listing of the item for sale in the marketplace (sellOrder == assetForSale)

decentraland/screens/MyAssetsScreen.js
marcelomorgado: I think so,
The source of that info is `CreatedOrder` events from Marketplace contract. So, each event has the assetId, type (nftAddress) and actionId (txhash).

pcowgill: Ah right, so there's not a separate lookup. Okay, makes sense! Thanks for the explanation. Resolving this.

decentraland/screens/MyAssetsScreen.js
pcowgill: But we're in the `My assets` screen for this comment thread. So the tx hash of the listing of the item for sale in the marketplace has nothing to do with the tx hash we'll want to use to build the Etherscan link on this screen. Does that mean there may be a bug for this feature?

decentraland/constants/UserActionStatus.js
pcowgill: Want to add this in in your current PR?

marcelomorgado: Done.

decentraland/screens/MyAssetsScreen.js
marcelomorgado: Oh, good question. I've misunderstood the terms here. I've used `sellOrders` wrongly.
Actually, assets from an address are being got using events too. We are using `Transfer` events to get that, so, the txhash applies here too.

pcowgill: Got it - that makes sense. Thanks! Resolving this.

decentraland/constants/UserActionStatus.js
pcowgill: Thanks! Resolving this

Issue #342, opened 2019-04-24
Open threads from PR #333


marcelomorgado:
[Rename `userAction` reducer](https://github.com/tasitlabs/tasit/pull/333#discussion_r277878914)
[Use keyed-object for `useraction`](https://github.com/tasitlabs/tasit/pull/333#discussion_r277883320)
[Change a message from info to warn](https://github.com/tasitlabs/tasit/pull/333#discussion_r277888035)

PR #343, opened 2019-04-24
Open threads from PR#333

marcelomorgado:
### Issue link

#342 (2 and 3)

### Types of changes

### DIFF
--- a/decentraland/App.js
+++ b/decentraland/App.js
@@ -108,7 +108,8 @@ Is the config file correct?`;
     // the introduction of userAction state (<= 0.0.17)
     myAssets.forEach(asset => {
       const { id: assetId, actionId } = asset;
-      let userAction = userActions.find(action => action.assetId === assetId);
+
+      let userAction = userActions[actionId];
 
       if (!userAction && !!actionId) {
         userAction = { [actionId]: { assetId, status: SUCCESSFUL } };
--- a/decentraland/App.js
+++ b/decentraland/App.js
@@ -17,13 +17,10 @@ import applyMiddleware from "./redux/middlewares";
 import { Action } from "tasit-sdk";
 const { ConfigLoader } = Action;
 import tasitSdkConfig from "./config/current";
-import {
-  checkBlockchain,
-  showFatalError,
-  restoreCreationStateOfAccountFromBlockchain,
-} from "@helpers";
+import { checkBlockchain, showFatalError } from "@helpers";
 import {
   retrieveAccount,
+  clearAllStorage,
   retrieveMyAssets,
   retrieveAccountCreationStatus,
   retrieveAccountCreationActions,
@@ -33,7 +30,6 @@ import {
 } from "@helpers/storage";
 import { Ionicons } from "@expo/vector-icons";
 import { Root } from "native-base";
-import { SUCCESSFUL } from "@constants/UserActionStatus";
 
 const store = createStore(decentralandApp, applyMiddleware);
 
@@ -71,51 +67,31 @@ Is the config file correct?`;
     // On iOS environment, Secure Store data remains even after app uninstallation
     const isFirstUse = await retrieveIsFirstUse();
     if (isFirstUse) {
-      // TODO: Uncomment this when 'isFirstUse' flag is being set widely
-      //await storeAccount(null);
+      await clearAllStorage();
       await storeIsFirstUse(false);
+      return;
     }
 
     const account = await retrieveAccount();
+    const creationStatus = await retrieveAccountCreationStatus();
+    const creationActions = await retrieveAccountCreationActions();
+
     if (account) {
       store.dispatch(setAccount(account));
-      let creationStatus = await retrieveAccountCreationStatus();
-      let creationActions = await retrieveAccountCreationActions();
+    }
 
-      const isOldAccount =
-        account !== null && creationStatus === null && creationActions === null;
+    if (creationStatus) {
+      store.dispatch(setAccountCreationStatus(creationStatus));
+    }
 
-      if (isOldAccount) {
-        ({
-          creationStatus,
-          creationActions,
-        } = await restoreCreationStateOfAccountFromBlockchain(account));
-      }
-
-      if (creationStatus !== null)
-        store.dispatch(setAccountCreationStatus(creationStatus));
-
-      if (creationActions !== null)
-        store.dispatch(setAccountCreationActions(creationActions));
+    if (creationActions) {
+      store.dispatch(setAccountCreationActions(creationActions));
     }
   }
 
   async _loadMyAssets() {
     const myAssets = await retrieveMyAssets();
-    let userActions = await retrieveUserActions();
-
-    // Handling with actions stored before
-    // the introduction of userAction state (<= 0.0.17)
-    myAssets.forEach(asset => {
-      const { id: assetId, actionId } = asset;
-
-      let userAction = userActions[actionId];
-
-      if (!userAction && !!actionId) {
-        userAction = { [actionId]: { assetId, status: SUCCESSFUL } };
-        userActions = { ...userActions, ...userAction };
-      }
-    });
+    const userActions = await retrieveUserActions();
 
     if (userActions) store.dispatch(addUserAction(userActions));
     if (myAssets) store.dispatch(setMyAssetsList(myAssets));
--- a/decentraland/App.js
+++ b/decentraland/App.js
@@ -113,7 +113,7 @@ Is the config file correct?`;
 
       if (!userAction && !!actionId) {
         userAction = { [actionId]: { assetId, status: SUCCESSFUL } };
-        userActions = { ...userActions, userAction };
+        userActions = { ...userActions, ...userAction };
       }
     });
 
--- a/decentraland/App.js
+++ b/decentraland/App.js
@@ -102,7 +102,7 @@ Is the config file correct?`;
 
   async _loadMyAssets() {
     const myAssets = await retrieveMyAssets();
-    const userActions = await retrieveUserActions();
+    let userActions = await retrieveUserActions();
 
     // Handling with actions stored before
     // the introduction of userAction state (<= 0.0.17)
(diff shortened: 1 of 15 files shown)


decentraland/screens/MyAssetsScreen.js
```diff
@@ -46,10 +46,12 @@ export class MyAssetsScreen extends React.Component {
...
     const assetsFromBlockchain = await this._getAssetsFromBlockchain(address);
 
-    const actionsFromBlockchain = assetsFromBlockchain.map(asset => {
+    let actionsFromBlockchain = {};
+
+    assetsFromBlockchain.forEach(asset => {
       const { actionId, id: assetId } = asset;
-      const status = SUCCESSFUL;
-      return { actionId, assetId, status };
+      const userAction = { [actionId]: { assetId, status: SUCCESSFUL } };
```
pcowgill: Why is `actionId` in `[]` brackets here?

decentraland/components/presentational/MyAssetsList.js
```diff
@@ -14,8 +14,15 @@ export default class MyAssetsList extends React.PureComponent {
     const { myAssets, userActions, renderItem } = this.props;
     const { length: listAmount } = myAssets;
 
+    const flatUserActions = Object.entries(userActions).map(userAction => {
+      const [actionId, userActionProps] = userAction;
```
pcowgill: Simple question to make sure I understand: What's the structure of userAction in this line? And then remind me what this syntax on the left does to it?

decentraland/screens/MyAssetsScreen.js
marcelomorgado: To use the `actionId` value as key:

`{ [actionId]: { assetId, status: SUCCESSFUL } };` => `{ '0x123...789': { assetId, status: SUCCESSFUL } };`
`{ actionId: { assetId, status: SUCCESSFUL } };` => `{ actionId: { assetId, status: SUCCESSFUL } };`

decentraland/components/presentational/MyAssetsList.js
marcelomorgado: Structure: `userActions` = `{ 0x123789: { assetId, status }, 0x987321: { assetId, status} }` same as `creationActions` but having txhash as key instead of status.
It's nice to get action info from an id: `info = userActions[actionId]` (O(1) complexity).
But to get an action for a given asset (since that no loger has the `actionId`) a find is needed, to iterate over the `userActions` to get the one related to an asset, I've used `Object.entries(userActions)`  that returns an array of arrays where the first item is a property and the second item is the value of that property.
So, for the example object above we'll have:
```javascript
Object.entries({ 0x123789: { assetId, status }, 0x987321: { assetId, status} }
//=> [ ['0x123789', { assetId, status }], ['0x987321', { assetId, status}] ]
```
The syntax used here is to assign arrays values to some props.
```javascript
const [actionId, userActionProps] = userAction;
// is the same as
const actionId = userAction[0];
const userActionProps = userAction[1];
```

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

decentraland/screens/MyAssetsScreen.js
pcowgill: Ah okay, I was guessing that might be it. Thanks for the explanation!

Resolving this.

decentraland/components/presentational/MyAssetsList.js
pcowgill: > same as `creationActions` but having txhash as key instead of status

Oh right, status for `creationActions` is essentially a nickname for what the tx is. Which is nice unless the action needs to be retried, but even then there could be two action objects under that status nickname.

> The syntax used here is to assign arrays values to some props.

Got it - I was familar with that syntax, but the part I wasn't getting was what had turned it into an array like that. So your explanation of Object.entries is what I was missing. That was very helpful, thanks!

**Other thoughts**

Maybe a little of this belongs as a comment in the code? Comments in the code that say at a low level what the code does are pointless, but comments in the code about why a choice was made and the goal behind it are very helpful at preventing introduction of bugs or unnecessary rewrites the next time someone looks at this section.

So maybe a comment that says:

```js
// It's easy to look up action info from actionId
// But to get an action for a given asset (since the asset itself no longer has the actionId)
// a find is needed, to iterate over the userActions to get the one related to an asset,
// we used Object.entries to turn it into an array of arrays to help with this find
``

File: decentraland/App.js [hunk 3]
```diff
@@ -113,7 +113,7 @@ Is the config file correct?`;
 
       if (!userAction && !!actionId) {
         userAction = { [actionId]: { assetId, status: SUCCESSFUL } };
-        userActions = { ...userActions, userAction };
+        userActions = { ...userActions, ...userAction };
```
pcowgill: Ahhh, interesting

decentraland/redux/middlewares.js
```diff
@@ -36,7 +36,7 @@ const storer = store => next => async action => {
       await storeAccountCreationStatus(creationStatus);
       break;
     }
-    case UPDATE_ACTION_FOR_ACCOUNT_CREATION_STATUS: {
+    case UPDATE_ACTION_ID_FOR_ACCOUNT_CREATION_STATUS: {
```
pcowgill: Good catch

decentraland/components/presentational/MyAssetsList.js
marcelomorgado: Done.

pcowgill: Thanks! Resolving this.

decentraland/redux/middlewares.js
pcowgill: Resolving this

decentraland/App.js

decentraland/App.js
```diff
@@ -71,51 +67,31 @@ Is the config file correct?`;
     // On iOS environment, Secure Store data remains even after app uninstallation
     const isFirstUse = await retrieveIsFirstUse();
     if (isFirstUse) {
-      // TODO: Uncomment this when 'isFirstUse' flag is being set widely
-      //await storeAccount(null);
+      await clearAllStorage();
       await storeIsFirstUse(false);
+      return;
```
pcowgill: Ah yeah, returning early here is important

decentraland/App.js
```diff
@@ -71,51 +67,31 @@ Is the config file correct?`;
...
-      const { id: assetId, actionId } = asset;
-
-      let userAction = userActions[actionId];
-
-      if (!userAction && !!actionId) {
-        userAction = { [actionId]: { assetId, status: SUCCESSFUL } };
-        userActions = { ...userActions, ...userAction };
-      }
-    });
+    const userActions = await retrieveUserActions();
```
pcowgill: Yeah, this is more readable with this extracted out. Nice.

decentraland/helpers/storage.js
```diff
@@ -1,3 +1,5 @@
+// Note: If a data wasn't found, returning null instead of default state of the
```
pcowgill: a data -> a piece of data
OR
a data -> data

decentraland/helpers/storage.js
```diff
@@ -1,3 +1,5 @@
...
+// property to avoid coupling with redux reducers
```
pcowgill: Good call

decentraland/helpers/storage.js
```diff
@@ -10,6 +12,15 @@ const EPHEMERAL_ACCOUNT_CREATION_ACTIONS = "EPHEMERAL_ACCOUNT_CREATION_ACTIONS";
 const IS_FIRST_APP_USE = "IS_FIRST_APP_USE";
 const USER_ACTIONS = "USER_ACTIONS";
 
+export const clearAllStorage = async () => {
+  await _clearData(IS_FIRST_APP_USE);
+  await _clearData(EPHEMERAL_ACCOUNT_PRIV_KEY);
```
pcowgill: I forget, are we just persisting the private key and not the address? And then we recover the address from the private key?

decentraland/helpers/storage.js
```diff
@@ -118,6 +129,15 @@ const _storeData = async (key, value, securely) => {
   }
 };
 
+const _clearData = async key => {
+  try {
+    await AsyncStorage.removeItem(key);
+    await SecureStore.deleteItemAsync(key);
```
pcowgill: Do we only want to do this if it failed to remove it from AsyncStorage? Or regardless. I guess the current approach is more defensive which is probably good.

decentraland/helpers/storage.js
marcelomorgado: piece of data, just commited

decentraland/App.js

decentraland/helpers/storage.js
marcelomorgado: Yes, we are persisiting the priv key:
```javascript
export const retrieveAccount = async () => {
  let account = null;

  const privateKey = await _retrieveData(EPHEMERAL_ACCOUNT_PRIV_KEY);
  if (privateKey != null) account = createFromPrivateKey(privateKey);

  return account;
};
```

decentraland/App.js

decentraland/helpers/storage.js
pcowgill: Cool, thanks! Resolving this

decentraland/helpers/storage.js
pcowgill: Resolving this

decentraland/helpers/storage.js
pcowgill: Cool, that's what I thought. Thanks. Resolving this.

decentraland/helpers/storage.js
marcelomorgado: Yes, I've tried to write a more defensive.
Your question makes me think if the behavior when failing, the docs aren't clear about that so I've changed this function to make sure that both functions will be executed.

marcelomorgado: Oh, that was already merged. I'll create a new PR.

pcowgill: > Yes, I've tried to write a more defensive.
> Your question makes me think if the behavior when failing, the docs aren't clear about that so I've changed this function to make sure that both functions will be executed.

Good thinking. Sounds good. Resolving this.

PR #345, opened 2019-04-24
Improve _clearData function

marcelomorgado:
### Issue link

https://github.com/tasitlabs/tasit/pull/343#discussion_r278281555

### Types of changes

### DIFF
--- a/decentraland/helpers/storage.js
+++ b/decentraland/helpers/storage.js
@@ -3,6 +3,7 @@
 import { SecureStore } from "expo";
 import { AsyncStorage } from "react-native";
 import { createFromPrivateKey } from "tasit-account/dist/testHelpers/helpers";
+import { logInfo } from "@helpers";
 
 // Storage keys
 const EPHEMERAL_ACCOUNT_PRIV_KEY = "EPHEMERAL_ACCOUNT_PRIV_KEY";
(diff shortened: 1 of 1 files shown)


decentraland/helpers/storage.js
```diff
@@ -132,9 +133,13 @@ const _storeData = async (key, value, securely) => {
 const _clearData = async key => {
   try {
     await AsyncStorage.removeItem(key);
+  } catch (error) {
+    logInfo(`Unable to delete data (key = ${key}) from storage.`);
```
pcowgill: `INFO` is a lower severity log than `WARN`. I would argue this is something that warrants a `WARN` level.

pcowgill: (Same for the other error for the secure storage)

pcowgill: Not important for this PR really, but mostly to set precedent for future code

marcelomorgado: Got it

marcelomorgado: I've made that way because if an exception is thrown for example if the key doesn't exist it'll log a warn on every clear attempt.

pcowgill: Oh, hmmm. That's an interesting point. So there is a pretty good argument for it being INFO, or splitting out based on whether that's the type of exception or if it's a different type of exception. But for now I'll merge this with how you have it, and we can ponder this going forward without there being an issue for it.

PR #353, opened 2019-04-25
Chaning back userAction from flat object to key-value object

### LINKED ISSUE #351: Purchased land Etherscan link using string “actionId” rather than the actionId

marcelomorgado:
### Issue link

#351 

### Auto-close the issue?

Closes #351 

### Types of changes

### DIFF
--- a/decentraland/components/presentational/MyAssetsList.js
+++ b/decentraland/components/presentational/MyAssetsList.js
@@ -24,9 +24,17 @@ export default class MyAssetsList extends React.PureComponent {
     });
 
     const dataList = myAssets.map(asset => {
-      let userAction = flatUserActions.find(
+      const flatUserAction = flatUserActions.find(
         action => action.assetId === asset.id
       );
+
+      let userAction;
+
+      if (flatUserAction) {
+        const { actionId } = flatUserAction;
+        userAction = { [actionId]: { ...userActions[actionId] } };
+      }
+
       return { asset, userAction };
     });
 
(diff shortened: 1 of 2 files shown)


pcowgill:
Thanks for the quick fix!

pcowgill:
@marcelomorgado I was going to merge this but CI is failing

pcowgill:
> @marcelomorgado I was going to merge this but CI is failing

I see you just fixed this!

marcelomorgado:
> @marcelomorgado Think users who have already run the last version of develop will have a smooth upgrade path?

Yes

pcowgill:
> Yes

On the data storage side, will there be anything like an extra key / value for the actions where the key is the string “actionId”? I haven’t through that carefully, but if that were the case, even if the user won’t see that, wondering if we want to clean that up to prevent uncaught exceptions later.

marcelomorgado:
This bug (accountId prop instead of '0x123' prop) was occured on the process that get data (`fromKey->toFlat->find`) from storage and parse it to the component.
The fix was to add an `->toKey` after the find.
So, no storage data was changed with this wrongly prop.

pcowgill:
> This bug (accountId prop instead of '0x123' prop) was occured on the process that get data (`fromKey->toFlat->find`) from storage and parse it to the component.
> The fix was to add an `->toKey` after the find.
> So, no storage data was changed with this wrongly prop.

Oh okay, that's great! Thanks for that explanation. Makes sense.

PR #357, opened 2019-04-25
0.0.19 -> 0.0.20

### LINKED ISSUE #356: Bump version and publish

marcelomorgado:
### Issue link
#356

### Auto-close the issue?

Closes #356


decentraland/package.json
```diff
@@ -1,19 +1,20 @@
 {
   "name": "tasit-decentraland",
-  "version": "0.0.19",
+  "version": "0.0.20",
   "description": "An app for buying Decentraland land using the Tasit SDK",
   "main": "node_modules/expo/AppEntry.js",
   "private": true,
   "scripts": {
+    "setup:testnet": "cp config/testnet.js config/current.js",
```
pcowgill: @marcelomorgado Do we need to add this? Doesn't `npm run start:testnet` do this already?

pcowgill: Oh, so it's using whatever `current.js` is on the developer's machine during the build process? I feel like it's weird I haven't encountered this error already myself, since I'm usually using the development config locally.

marcelomorgado: 1. Yes, it does
2. The `App.js` will load the config from `config/current.js` so my intent is to guarantee that this file is pointing to testnet before running build/publish scripts

marcelomorgado: Is that makes sense?

pcowgill: Yep, that makes sense - I'm just perplexed if it wasn't working that way already how we haven't run into this before. It's probably because the old npm start used to default to testnet

PR #467, opened 2019-12-17
Deploying Safe contract in-app, ts cleanup, removed deprecated, scripts cleanup

pcowgill:
### Types of changes


decentraland/helpers/index.ts
```diff
@@ -155,25 +256,31 @@ export const fundAccountWithMana = (accountAddress): ActionObject => {
...
+const showToast = (msg: string): void =>
   Toast.show({ text: msg, duration: 3000, buttonText: "Okay" });
 
-export const showFatalError = (msg): void => console.error(msg);
-export const showError = (msg): void => showToast(`ERROR: ${msg}`);
-export const showWarn = (msg): void => showToast(`WARN: ${msg}`);
-export const showInfo = (msg): void => showToast(`${msg}`);
+export const showFatalError = (msg: string): void => console.error(msg);
+export const showError = (msg: string): void => showToast(`ERROR: ${msg}`);
+export const showWarn = (msg: any): void => showToast(`WARN: ${msg}`);
```
pedrobergamini, line 278: I think here msg could be annotated as string
```suggestion
export const showWarn = (msg: string): void => showToast(`WARN: ${msg}`);
```

decentraland/components/stateful/EthereumUpgradeSecurity/EthereumUpgradeSecurity.tsx
```diff
@@ -2,38 +2,330 @@ import React, { useState, useEffect } from "react";
...
-    handleDeploymentComplete();
+    onCallSafeContractSubmit();
 
     return function cleanup() {
       // TODO: Cancel any unwanted remnants of having already made this request
     };
-  });
+  }, [account]); // Only re-run the effect if account changes
+
+  // TODO: Show loading by setting a boolean or enum value when deployment starts
```
pedrobergamini, line 293: Maybe we could use a state hook for that to set loading value true/false inside `onCallSafeContractSubmit` try/catch block

pedrobergamini:
Very nice! Only suggested a few small changes, what do you think?

decentraland/components/stateful/EthereumUpgradeSecurity/EthereumUpgradeSecurity.tsx
pcowgill: @pedrobergamini Yep, I think that sounds good!

pcowgill: How would that handle persistence if someone quits the app and reopens it during the process?

pcowgill: I created an issue to track this improvement. https://github.com/tasitlabs/tasit/issues/478

